PR #62, opened 2018-01-02
Add md5 hash as additional return value for all services

HendrikLeuschner:
md5 hash extracted from any single .md5 file with any name in folder 'graphs' and returned as parameter in response of all services.


TimMcCauley:
@HendrikLeuschner Is the md5 file read for every request?

TimMcCauley (changes requested):
Should be read on initialize once and help globally for every service to read.

openrouteservice/src/main/java/heigit/ors/config/AppConfig.java
```diff
@@ -51,6 +55,21 @@ public AppConfig()	{
...
+			public boolean accept(File pathname) {
+				return pathname.getName().endsWith(".md5");
+			}
+		});
+		if (md5Files.length == 1){
+			try{
+				osm_md5_hash = FileUtility.readFile(md5Files[0].toString()).trim();
+			}
+			catch (IOException e)
+			{}
```
rabidllama: It might be worth adding an error message to the log if there is a problem reading the md5 files seeing as this is a new feature and so would be easier for debugging. It doesn't need to be sent to the user, just logged.

rabidllama (changes requested):

rabidllama (approved):

TimMcCauley (approved):

PR #73, opened 2018-01-10
Remove borders code

rabidllama:
Removed all code that was added as part of the borders feature. This will be added as its own branch


TimMcCauley (approved):
OK

Issue #70, opened 2018-01-09
Replace import sun.misc.Base64Encoder


sfendrich:
The classes in sun.* are not portable across different Java implementations, e.g., "mvn compile" fails on my machine due to heigit/ors/routing/traffic/TrafficUtility.java importing sun.misc.Base64Encoder (see [Why Developers Should Not Write Programs That Call 'sun' Packages](http://www.oracle.com/technetwork/java/faq-sun-packages-142232.html)). java.util.Base64 would be a better choice.

TimMcCauley:
@sfendrich thanks for pointing this out - feel free to change and push to the repository

PR #74, opened 2018-01-10
Fix issue #70

sfendrich:
Packages in sun.* are not portable across different Java implementations. Better use java.util.Base64 instead of sun.misc.Base64Encoder


rabidllama (approved):
Do we know if the password is an actual password or just a dummy one?

PR #76, opened 2018-01-12
Fixed issue with hgv restrictions only working when certain number of…

### LINKED ISSUE #75: HGV restrictions not working correctly
It seems that HGV restrictions (at least width) are only taken into account if other restrictions are also added. 

The following way has a maxwidth property of 2
http://www.openstreetmap.org/way/37855163

When a restriction is set on hgv routing for this it is NOT taken into account

https://openrouteservice.org/directions?n1=49.428983&n2=8.689633&n3=17&a=49.430117,8.690915,49.427758,8.68834&b=4a&c=1&f2=3&k1=en-US&k2=km

when other restrictions are added that shouldn't affect the route, the width IS taken into account, but only when a minimum of 3 restrictions are added

https://openrouteservice.org/directions?n1=49.428236&n2=8.690073&n3=17&a=49.430117,8.690915,49.427758,8.68834&b=4a&c=1&f2=3&f5=2&f3=2&k1=en-US&k2=km

rabidllama:
… restrictions are entered. Fixes #75


TimMcCauley (approved):
OK

PR #82, opened 2018-01-23
Borders avoid

rabidllama:
Added feature to avoid border crossings using the same mechanism as AvoidFeatures (EdgeFilters)


TimMcCauley (approved):
OK

PR #84, opened 2018-01-23
GPXBuild Patch

MichaelsJP:
Set Config Parameter Optional
Added a new Exceptionclass that can be used to provide information about missing App.config parameter

### DIFF
--- a/openrouteservice/src/main/java/heigit/ors/exceptions/MissingConfigParameterException.java
+++ b/openrouteservice/src/main/java/heigit/ors/exceptions/MissingConfigParameterException.java
@@ -1,0 +1,8 @@
+package heigit.ors.exceptions;
+
+public class MissingConfigParameterException {
+
+    public MissingConfigParameterException(String missingConfigParameter) {
+        System.err.print("Config parameter " + missingConfigParameter + " is missing." + "\n");
+    }
+}
--- a/openrouteservice/src/main/java/heigit/ors/exceptions/MissingConfigParameterException.java
+++ b/openrouteservice/src/main/java/heigit/ors/exceptions/MissingConfigParameterException.java
@@ -1,5 +1,6 @@
 package heigit.ors.exceptions;
 
+import heigit.ors.services.routing.requestprocessors.gpx.GpxRoutingResponseWriter;
 import org.apache.log4j.Logger;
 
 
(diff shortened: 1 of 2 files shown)


openrouteservice/src/main/java/heigit/ors/exceptions/MissingConfigParameterException.java
```diff
@@ -0,0 +1,8 @@
+package heigit.ors.exceptions;
+
+public class MissingConfigParameterException {
+
+    public MissingConfigParameterException(String missingConfigParameter) {
+        System.err.print("Config parameter " + missingConfigParameter + " is missing." + "\n");
```
rabidllama: Would be better if this used the Log4j logging instead of standard System printing. (See ors/config/AppConfig.java for an example)

openrouteservice/src/main/java/heigit/ors/services/routing/requestprocessors/gpx/GpxRoutingResponseWriter.java
```diff
@@ -129,26 +132,70 @@ public static String toGPX(RoutingRequest rreq, RouteResult[] routeResults) thro
...
+        // set content_licence
+        if (AppConfig.Global().getParameter("info", "content_licence") != null) {
+            copyright.setLicense(AppConfig.Global().getParameter("info", "content_licence"));
+        } else {
+            copyright.setLicense("");
+            new MissingConfigParameterException("content_licence");
+        }
         // create and set current date as XMLGregorianCalendar element
         Date date = new Date();
         GregorianCalendar c = new GregorianCalendar();
```
rabidllama: Can we not use general java date here rather than a new object? `int year = Calendar.getInstance().get(Calendar.YEAR);` should work as far as I know

rabidllama (changes requested):
Generally looks OK, just a couple of things that would make it a bit more sustainable relating to logging and date objects

openrouteservice/src/main/java/heigit/ors/services/routing/requestprocessors/gpx/GpxRoutingResponseWriter.java
rabidllama: Ignore this for now - we can update later if needed

openrouteservice/src/main/java/heigit/ors/exceptions/MissingConfigParameterException.java
MichaelsJP: I moved the error message handling to Log4j logging using LOGGING.info except one LOGGING.error for a malformed mail

openrouteservice/src/main/java/heigit/ors/exceptions/MissingConfigParameterException.java
```diff
@@ -1,5 +1,6 @@
 package heigit.ors.exceptions;
 
+import heigit.ors.services.routing.requestprocessors.gpx.GpxRoutingResponseWriter;
```
rabidllama: Can we ditch this import?

rabidllama (approved):
OK

Issue #63, opened 2018-01-02
Potential matrix issues


karussell:
I'm relatively certain that they are some bugs but I'm not 100% sure as I have integrated the code differently using an unmodified GraphHopper.

You can try to reproduce them if you run a few (5000) matrix request with 2 or 5 points over berlin area via comparing with single 1 to 1 requests.

> [RPHASTAlgorithm.calcPaths](https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/heigit/ors/routing/algorithms/RPHASTAlgorithm.java#L181)

```java
 for (int i = 0; i < sourceIds.length; i++) {
  int sourceNode = sourceIds[i];
  if (sourceNode == -1)
     continue;

  // BUG FIX 1
  MultiTreeSPEntry existing = _bestWeightMapFrom.get(sourceNode);
  if (existing != null) {
     existing.getItem(i).weight = 0.0001;
     continue;
  }

  ...
}
```

and

```java
// BUG FIX 2: if query node is directly part of the subgraph (no virtual node) it could be higher than a leaf from the highest node and not reachable in runDownwardSearch
// TODO write test where query node is 6763 and shortest is over 6763->35469->47681 but only 6763-5422-47681 is returned without the fix as 6763 is not reached from 5422
//         .-> 5422 ------
// 6763 <-/               \
//    \------> 35469 <-\   \
//                      -- 47681
for (int i = 0; i < sourceIds.length; i++) {
     int sourceNode = sourceIds[i];
     MultiTreeSPEntry mspTree = _bestWeightMapFrom.get(sourceNode);
     mspTree.getItem(i).update = true;
      _prioQueue.add(mspTree);
}

 _outEdgeExplorer = _targetGraph.createExplorer();
runDownwardSearch();
```

And finally a more tricky one in [RPHASTAlgorithm.fillEdgesDownward](https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/heigit/ors/routing/algorithms/RPHASTAlgorithm.java#L345) where update=false is hindering a necessary update of a better path - I've just commented out the continue, which only slightly decreases query speed. Currently unsure why the upward search "continue" does not lead to problems.

```java

// BUG FIX 3: this optimization leads to suboptimal routes in rare cases of bigger matrices (berlin, measurement.count=30 measurement.points_per_query=5)
// if (_msptItem.update == false) {
//    continue;
// }
```

[...]

PR #86, opened 2018-01-25
Matrix bug fixes and API test

HendrikLeuschner:
fixes issues of #63 and adds an api test for matrix

### DIFF
--- a/openrouteservice-api-tests/src/test/java/heigit/ors/services/matrix/ResultsValidationTest.java
+++ b/openrouteservice-api-tests/src/test/java/heigit/ors/services/matrix/ResultsValidationTest.java
@@ -212,9 +212,15 @@ public class ResultsValidationTest extends ServiceTest {
         Assert.assertEquals(jTable.length(), rows);
         Assert.assertEquals(jTable.getJSONArray(0).length(), columns);
     }
-
+    /**
+     * Queries the matrix API with 12x12 symmetrical matrix. Queries the routing API
+     * with the same 12x12 single queries. Compares results. If results are within .2m of
+     * each other, test passes. This way the result of the matrix API is bound to be
+     * the same as the routing API
+     */
     @Test
     public void distanceTest() {
+        //Query Matrix API
         Response response = given()
                 .param("locations", getParameter("manyLocations"))
                 .param("sources", "all")
@@ -226,11 +232,11 @@ public class ResultsValidationTest extends ServiceTest {
                 .get(getEndPointName());
 
         String[] locations = (String[]) getParameter("manyLocationsArray");
-
-
+        
         Assert.assertEquals(response.getStatusCode(), 200);
         JSONObject jResponse = new JSONObject(response.body().asString());
         JSONArray jDistances = jResponse.getJSONArray("distances");
+        //Query Routing API 12x12 times
         for(int i = 0; i < 12; i++) {
             for (int j = 0; j < 12; j++) {
                 Response response2 = given()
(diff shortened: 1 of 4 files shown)


openrouteservice/src/main/java/heigit/ors/routing/algorithms/RPHASTAlgorithm.java
```diff
@@ -275,19 +290,18 @@ private void fillEdgesUpward(MultiTreeSPEntry currEdge, PriorityQueue<MultiTreeS
...
-		//					upwardItersSkipped++;
+							//					upwardItersSkipped++;
 							continue;
 						}
+						_msptSubItem = ee.getItem(i);
 						
 						_tmpWeight = _edgeWeight + _entryWeight;
 
-						if (_msptSubItem.weight > _tmpWeight || _msptSubItem.weight == 0.0) {
+						if (_msptSubItem.weight > _tmpWeight || _msptSubItem.weight == Double.POSITIVE_INFINITY) {
```
karussell: I think just `_msptSubItem.weight > _tmpWeight` is necessary

openrouteservice/src/main/java/heigit/ors/routing/algorithms/RPHASTAlgorithm.java
```diff
@@ -275,19 +290,18 @@ private void fillEdgesUpward(MultiTreeSPEntry currEdge, PriorityQueue<MultiTreeS
 						_msptItem = currEdge.getItem(i);
 						_entryWeight = _msptItem.weight;
 						
-						if (_entryWeight == 0.0)
+						if (_entryWeight == Double.POSITIVE_INFINITY)
```
karussell, line 293: You should use Double.isInfinity instead

HendrikLeuschner: If you look at the implementation of Double.isInfinite, it its
```public static boolean isInfinite(double var0) {
        return var0 == 1.0D / 0.0 || var0 == -1.0D / 0.0;
 }
```
of which 50% is the definition of Double.POSITIVE_INFINITY:
```
public static final double POSITIVE_INFINITY = 1.0D / 0.0;
```
As I do not want to accept Double.NEGATIVE_INFINITY, which would hint at an error in my code, I am happy to use my version, which should be faster as well.

openrouteservice/src/main/java/heigit/ors/routing/algorithms/RPHASTAlgorithm.java
HendrikLeuschner: True!

openrouteservice-api-tests/src/test/java/heigit/ors/services/matrix/ResultsValidationTest.java
```diff
@@ -89,95 +122,139 @@ public void emptyLocationEntryTest() {
...
+    private void checkTableDimensions(JSONObject json, String tableName, int rows, int columns) {
+        Assert.assertEquals(true, json.has(tableName));
+
+        JSONArray jTable = json.getJSONArray(tableName);
+        Assert.assertEquals(jTable.length(), rows);
+        Assert.assertEquals(jTable.getJSONArray(0).length(), columns);
+    }
+
+    @Test
+    public void distanceTest() {
```
rabidllama: Can we write a bit of documentation about what the test is for?

openrouteservice-api-tests/src/test/java/heigit/ors/services/matrix/ResultsValidationTest.java
```diff
@@ -89,95 +122,139 @@ public void emptyLocationEntryTest() {
...
+
+                JSONObject jResponseRouting = new JSONObject(response2.body().asString());
+                JSONObject jRoute = (jResponseRouting.getJSONArray("routes")).getJSONObject(0);
+                double routeDistance = jRoute.getJSONObject("summary").getDouble("distance");
+//                double matrixDistance;
+
+                double matrixDistance = jDistances.getJSONArray(i).getDouble(j);
+                Assert.assertTrue( matrixDistance - .1 < routeDistance);
+                Assert.assertTrue( matrixDistance + .1 > routeDistance);
+//                        .assertThat()
```
rabidllama: We should try and make sure that we don't leave commented out code when we push to dev/master

openrouteservice-api-tests/src/test/java/heigit/ors/services/matrix/ResultsValidationTest.java
```diff
@@ -21,6 +21,7 @@
 package heigit.ors.services.matrix;
 
 import static io.restassured.RestAssured.*;
+import static org.hamcrest.Matchers.*;
```
rabidllama: Do any objects from this package get used? If not then it shouldn't be imported.

rabidllama (changes requested):
Code seems OK, but we need to make sure that any new methods created which are complicated (especially tests) are well documented.

rabidllama (approved):

PR #102, opened 2018-02-02
Matrix flexible mode location count restriction

### LINKED ISSUE #94: Restrict Matrix locations count when using Dijkstra, configurable via app.config

HendrikLeuschner:
Restricts the matrix requests locations to a diffferent (lower) value if not using profile driving-car or hgv (i.e. if there is no CH profile for the requested profile).
The additional parameter still needs to be added to the app.config in
``` ors -> services -> matrix ```
after the entry ```maximum_locations``` and is called ```maximum_locations_flexible: 25 ``` @TimMcCauley please 😄 
Fallback to value of 25 if the parameter is not set is hardcoded.
Resolves #94

### DIFF
--- a/openrouteservice/src/main/java/heigit/ors/services/matrix/MatrixServiceSettings.java
+++ b/openrouteservice/src/main/java/heigit/ors/services/matrix/MatrixServiceSettings.java
@@ -25,6 +25,7 @@ import heigit.ors.config.AppConfig;
 public class MatrixServiceSettings 
 {
 	private static int maximumLocations = 100;
+	private static int maximumLocationsFlexible = 25;
 	private static int maximumVisitedNodes = 100000;
 	private static double maximumSearchRadius = 2000;
 	private static boolean allowResolveLocations = true;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: openrouteservice/src/main/java/heigit/ors/services/matrix/MatrixServiceSettings.java [hunk 1]
```diff
@@ -39,6 +40,9 @@
 		value = AppConfig.Global().getServiceParameter("matrix", "maximum_locations");
 		if (value != null)
 			maximumLocations = Math.max(1, Integer.parseInt(value));
+		value = AppConfig.Global().getServiceParameter("matrix", "maximum_locations_flexible");
```
rabidllama, line 43: I have added the parameter to the staging app.config, but it would be good to have it added to the app.config.sample file within the repository so that it is there ready for new instances

rabidllama (changes requested):
Just needs the maximum_locations_flexible property adding to the app.config.sample file along with a comment about what it is for so that people creating future instances know that it is available.

Also, please add a little about it to the CHANGELOG.md under the [Unreleased] heading.

rabidllama (approved):

PR #129, opened 2018-02-27
Release 4.5

rabidllama:
Upgrade from 4.4.2 to 4.5.

Changes include (from changelog.md):
### Added
- Functionality has been added to restrict routes so that they do not cross all borders, controlled borders, or the borders of specific countries
- Added GeoJson export for routing exports
- Added global export class to combine all exports there

### Fixed
- Fix exception when roundabout exit is not correctly found
- Option to specify maximum locations for matrix request when using non-standard weightings
- Geocoder now returns a 404 response if no address is found for reverse geocoding
- Fixed error codes
- Correct querying of population statistics data for isochrones

### Changed
- RoutingProfile was changed to make sure whenever pop_total or pop_area is queried, both are present in the attributes
- Response with a detour factor now uses "detourfactor" rather than "detour_factor"
- Changed the gpx export to the new global export processor

### Deprecated
- getStatisticsOld | Connected to the old statistics library
- geometryToWKB | Connected to the old statistics library


TimMcCauley (approved):
OK

PR #179, opened 2018-05-24
removing the code that is merging turn instructions - Issue #177

### LINKED ISSUE #177: RouteResultBuilder Issue
Hi,

as I started in the google groups 
https://groups.google.com/d/msg/openrouteservice/o5-ZTUZVP-I/vU8C2rvuBwAJ

I believe there is an issue with the route instruction builder... I have collected few more examples:

Here BOTH of the roundabout turning instructions will be removed

https://maps.openrouteservice.org/directions?n1=51.830152&n2=8.446485&n3=17&a=51.831362,8.448905,51.828949,8.444045&b=0&c=0&k1=en-US&k2=km

while here

https://maps.openrouteservice.org/directions?n1=51.830152&n2=8.446485&n3=17&a=51.831362,8.448905,51.828995,8.445407&b=0&c=0&k1=en-US&k2=km

they will be keept...

I am not 100% sure, but I believe that the root cause of this issue can be found in the heigit.ors.routing.RouteResultBuilder - starting from line 220

`// merge route steps with similar names 
`
here TurnInstructions of the same type will be merged, if the "name1" contains "name2" or vice versa - in the code there is an example URL which should demonstrate why this code can be useful -  unfortunately I am not able to check/understand this route [it's quite long] - so might be the code is removing some turn instructions with a good reason - but the check

`
if (prevStep != null && instrType == prevInstrType && canMergeInstructions(instr.getName(), prevInstr.getName()))
`
is IMHO not correct (or way to less) - in my initial case two right turns will be merged into one - or in my second example the roundabout turn instructions will be merged...

So IF this merging is required (in the given example URL) then you should add additional checks for specific turn types ONLY - and second check which might be helpful is the distance between the the "possible merge targets" ?

Thanks for spending time with this issue report - if you want/need more examples - please let me know ;-)

Cheers
Matthias

marq24:
### Pull Request Checklist
- [x ] 1. I have merged the latest version of the development branch into my feature branch and all conflicts have been resolved
- [ -] 2. I have added information about the change/addition to functionality to the CHANGELOG.md file under the [Unreleased] heading
- [ -] 3. I have documented my code using JDocs tags
- [ x] 4. I have removed unecessary commented out code, imports and System.out.println statements
- [ -] 5. I have written JUnit tests for any new methods/classes and ensured that they pass
- [ -] 6. I have created API tests for any new functionality exposed to the API
- [ -] 7. If changes/additions are made to the app.config file, I have added these to the app.config.sample file along with a short description of what it is for, and documented this in the Pull Request (below) 
- [ x] 8. I have built graphs with my code of the Heidelberg.osm.gz file and run the api-tests with all test passing
**-> there are 2 TESTS that are failing right now! [ the GeoJSON Tests that are checking for 'features' = yes] - i have no clue what need to be done so that they will pass - but this is already failing with the original development branch**

- [ x] 9. I have referenced the Issue Number in the Pull Request (if the changes were from an issue)
- [ -] 10. For new features involving building of graphs, I have tested on a larger dataset (at least Germany) and the graphs build without problems (i.e. no out-of-memory errors)
- [ x] 11. I have written in the Pull Request information about the changes made including their intended usage and why the change was needed.

Fixes #177 .

- Key functionality added:
well actually I have removed the TurnInstruction Merging code - simply cause I have found only cases where REQUIRED turn instructions will be removed.

- Reason for change: 
Required Turn Instructions will be removed from the Instruction list (merged) that should not be removed - actually might be that the code was required in the past - but honestly I already missed two times my route (on my road bike) cause if this turn instruction "optimization"... And IF there was ever any use - then the negative impact is way to massive.

- none

Kind Regards
Matthias

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Fixed problem with avoid polygons excluding ways that should have been accepted (Issue #95)
+- Updated code to remove merging of instructions as this resulted in missing important turn instructions (Issue #177)
 
 ### Changed
 -
--- a/openrouteservice-api-tests/src/test/java/heigit/ors/services/routing/ResultTest.java
+++ b/openrouteservice-api-tests/src/test/java/heigit/ors/services/routing/ResultTest.java
@@ -568,7 +568,7 @@ public class ResultTest extends ServiceTest {
 				.body("routes[0].segments[0].containsKey('steps')", is(true))
 				.body("routes[0].segments[1].containsKey('steps')", is(true))
 				.body("routes[0].segments[0].steps.size()", is(55))
-				.body("routes[0].segments[1].steps.size()", is(31))
+				.body("routes[0].segments[1].steps.size()", is(33))
 				.statusCode(200);
 	}
 
@@ -588,7 +588,7 @@ public class ResultTest extends ServiceTest {
 				.body("routes[0].segments[0].containsKey('steps')", is(true))
 				.body("routes[0].segments[1].containsKey('steps')", is(true))
 				.body("routes[0].segments[0].steps.size()", is(55))
-				.body("routes[0].segments[1].steps.size()", is(31))
+				.body("routes[0].segments[1].steps.size()", is(33))
 				.body("routes[0].segments[0].steps[0].distance", is(511.4f))
 				.body("routes[0].segments[0].steps[0].duration", is(230.1f))
 				.body("routes[0].segments[0].steps[0].type", is(11))
--- a/openrouteservice/src/main/java/heigit/ors/routing/RouteResultBuilder.java
+++ b/openrouteservice/src/main/java/heigit/ors/routing/RouteResultBuilder.java
@@ -130,7 +130,7 @@ public class RouteResultBuilder
 					//	nInstructions -= 1;
 
 					Instruction instr, prevInstr = null;
-					InstructionType instrType, prevInstrType = InstructionType.UNKNOWN;
+					InstructionType instrType = InstructionType.UNKNOWN;
 					RouteSegment seg = new RouteSegment(path, units);
 
 					if (includeDetourFactor)
@@ -145,7 +145,6 @@ public class RouteResultBuilder
 						seg.setDetourFactor((dist == 0) ? 0 : FormatUtility.roundToDecimals(path.getDistance()/dist, 2));
 					}
 
-					RouteStep prevStep = null;
 					String instrText = "";
 					double stepDistance, stepDuration;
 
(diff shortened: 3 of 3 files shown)


rabidllama:
I agree that this is definitely an issue in terms of missing out important instructions. I believe that the original logic was to try and cut down the number of instructions that would give the same information (i.e. on a large curved bend, you may get multiple instructions saying "turn slight right").

I think that we should look into this again at some point and try to re-include the merging of instructions where it is suitable, as in the old code the logic seems a little amiss.

nonmerge_correct
merge_incorrect

The above two images show where merging of instructions is not valid.

The two images below shows where the merging of instructions could be seen as better

merge_correct
unmerge_incorrect

rabidllama (changes requested):
I have copied the changed code into a local version of ORS on my machine, and two of the API tests are failing (ResultTest.testSteps:571 and ResultTest.testStepsDetails:591).

Can you update both of those tests so that they pass as they are both related to the number of steps (and so instructions) that are returned in the response. 

To run the API tests, you need to copy the app.config.test file from the openrouteservice-api-tests/conf folder into the openrouteservice/WebContent/WEB-INF folder and use that for building ORS (change its name to app.config, but make sure that you rename your current app.config to something like app.config.dev so you don't lose your changes in there) and then rebuild and run the ORS service. Then in the openrouteservice-api-tests run "mvn clean install" to run the api tests. 

If there are any problems let me know and I will try to help.

marq24:
I am working on the test fixes asap...

for the given example where it "might" had been sense full that turn instructions will be merged - 

http://localhost:3035/directions?n1=49.418674&n2=8.692581&n3=18&a=49.417653,8.691489,49.42015,8.689891&b=0&c=0&k1=en-US&k2=km

from my perspective the answer would be "Yes & No" - IMHO this is an issue of the graph hopper turn instructions itself - I see them at many places - specially when street names are changing (and even if there is no junction in the OSM data)...

So WHEN you want to get rid of "unnecessary" turn instructions, then this should/have to be addressed within GraphHopper - since there we have the information about the junction/nodes (and the possible alternative ways)...

In the Turn instruction the infos are already lost (what are the alternative options at a certain node) - and with that it's IMHO impossible to make the decision about sense or nonesense.

rabidllama (approved):

rabidllama:
tests all pass now thanks - I am not sure why you were having a problem with the GeoJSON test as it worked ok on my local version on both the development branch and on the one with the above changes...

rabidllama:
@marq24 sorry, one last thing that I missed in the original review - can you add the following to the CHANGELOG.md (under [Unreleased] Fixed:) please so that everything is all kept together in the one pull request:

Updated code to remove merging of instructions as this resulted in missing important turn instructions (Issue #177)

marq24:
I hope now all is complete and ready to merge :-)

rabidllama:
All merged into development, thanks

PR #356, opened 2018-11-15
Api refactoring matrix

MichaelsJP:
### Pull Request Checklist

- [ ] 1. I have merged the latest version of the development branch into my feature branch and all conflicts have been resolved.

Fixes # .

- Key functionality added:
- Reason for change: 

- 

-


openrouteservice/src/main/java/heigit/ors/api/controllers/MatrixAPI.java
```diff
@@ -0,0 +1,94 @@
...
+import heigit.ors.matrix.MatrixErrorCodes;
+import heigit.ors.matrix.MatrixResult;
+import io.swagger.annotations.*;
+import org.springframework.http.ResponseEntity;
+import org.springframework.http.converter.HttpMessageNotReadableException;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.List;
+
+import static heigit.ors.api.requests.matrix.MatrixRequestHandler.convertMatrixRequest;
```
rabidllama: I would personally not import a static method and instead have it in the main part of the code as it can make it difficult to follow where things are going.
So the import is import ......matrix.MatrixRequestHandler;
and then in the code MatrixRequestHandler.convertMatrixRequest(...);

openrouteservice-api-tests/src/test/java/heigit/ors/v2/services/config/AppConfig.java
```diff
@@ -0,0 +1,156 @@
+/*  This file is part of Openrouteservice.
```
rabidllama, line 1: Would it be possible to move this outside of the v2 package as it is not really dependant on the api version.

openrouteservice-api-tests/src/test/java/heigit/ors/v2/services/config/AppConfig.java
```diff
@@ -0,0 +1,156 @@
...
+ *  This library is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *  See the GNU Lesser General Public License for more details.
+
+ *  You should have received a copy of the GNU Lesser General Public License along with this library;
+ *  if not, see <https://www.gnu.org/licenses/>.
+ */
+package heigit.ors.v2.services.config;
+
+import com.typesafe.config.*;
```
rabidllama, line 16: I think we changed all app.onfigs to be only JSON now, so it would make sense to use JSON rather than typesafe. @TimMcCauley is this the case?

openrouteservice-api-tests/src/test/java/heigit/ors/v2/services/config/RoutingProfileType.java
```diff
@@ -0,0 +1,251 @@
...
+    public static final int DRIVING_TRAFFIC = 8;
+
+    // CYCLING STUFF
+    public static final int CYCLING_REGULAR = 10;
+    public static final int CYCLING_MOUNTAIN = 11;
+    public static final int CYCLING_ROAD = 12;
+    public static final int CYCLING_ELECTRIC = 17;
+
+    public static final int CYCLING_MOTOR = 15;
+    @Deprecated
```
rabidllama: We should remove the deprecated items from here as they are not present anymore

openrouteservice-api-tests/src/test/java/heigit/ors/v2/services/config/RoutingProfileType.java
```diff
@@ -0,0 +1,251 @@
...
+            case CYCLING_SAFE:
+                return "cycling-safe";
+            case CYCLING_TOUR:
+                return "cycling-tour";
+            case CYCLING_ELECTRIC:
+                return "cycling-electric";
+
+            case FOOT_WALKING:
+                return "foot-walking";
+            case FOOT_WALKING_DO_NOT_USE:
```
rabidllama: the *_DO_NOT_USE ones should not be included here

openrouteservice-api-tests/src/test/java/heigit/ors/v2/services/serviceSettings/IsochronesServiceSettings.java
```diff
@@ -0,0 +1,187 @@
...
+
+    public static int getMaximumLocations() {
+        return maximumLocations;
+    }
+
+    public static int getMaximumIntervals() {
+        return maximumIntervals;
+    }
+
+    /*public static int getMaximumRange(int profileType, TravelRangeType range) {
```
rabidllama: If not used, then delete it

openrouteservice/src/main/java/heigit/ors/api/converters/APIMatrixRequestProfileConverter.java
```diff
@@ -0,0 +1,31 @@
...
+ * You should have received a copy of the GNU Lesser General Public License along with this library;
+ * if not, see <https://www.gnu.org/licenses/>.
+ */
+
+package heigit.ors.api.converters;
+
+import heigit.ors.api.requests.common.APIEnums;
+import org.springframework.core.convert.converter.Converter;
+
+public class APIMatrixRequestProfileConverter implements Converter<String, APIEnums.MatrixProfile> {
```
rabidllama: We should rename the RoutingProfile enum to be Profile and then we can use that for the matrix and isochrones aswell. That would mean that we would not need this extra converter and the extra Enum

openrouteservice/src/main/java/heigit/ors/api/converters/APIRoutingRequestProfileConverter.java
```diff
@@ -18,7 +18,7 @@
 import org.springframework.core.convert.converter.Converter;
 import heigit.ors.api.requests.common.APIEnums;
 
-public class APIRequestProfileConverter implements Converter<String, APIEnums.RoutingProfile> {
+public class APIRoutingRequestProfileConverter implements Converter<String, APIEnums.RoutingProfile> {
```
rabidllama: See above - if we update the enums then we do not need to rename this converter and it can be reused for the matrix

openrouteservice/src/main/java/heigit/ors/api/requests/common/APIEnums.java
```diff
@@ -255,8 +272,43 @@ public String toString() {
...
+        }
+
+        @Override
+        @JsonValue
+        public String toString() {
+            return value;
+        }
+    }
+
+    public enum MatrixProfile {
```
rabidllama: If we rename the RoutingProfile enum then we can use that for matrix aswell and so we dont need to create this extra one

openrouteservice/src/main/java/heigit/ors/api/requests/matrix/MatrixRequest.java
```diff
@@ -0,0 +1,232 @@
...
+    @ApiModelProperty(name = "resolve_locations", value = "Specifies whether given locations are resolved or not. If the parameter value set to `true`, every element in destinations and sources will contain `name` element that identifies the name of the closest street. Default is `false`")
+    @JsonProperty(value = "resolve_locations", defaultValue = "false")
+    private Boolean resolveLocations = false;
+
+    @ApiModelProperty(name = "units", value = "Specifies the distance unit.\n" +
+            "Default: m.")
+    @JsonProperty(value = "units", defaultValue = "m")
+    private String units = "m";
+
+    @ApiModelProperty(name = "optimized", value = "Specifies weather flexible mode is used or not.", hidden = true)
```
rabidllama: whether, not weather

openrouteservice/src/main/java/heigit/ors/api/responses/matrix/JSONMatrixResponseObjects/JSONIndividualMatrixResponse.java
```diff
@@ -0,0 +1,150 @@
...
+import heigit.ors.matrix.MatrixRequest;
+import heigit.ors.matrix.MatrixResult;
+import heigit.ors.util.FormatUtility;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+@ApiModel(value = "JSONIndividualRouteResponse", description = "An individual JSON based route created by the service")
```
rabidllama: Needs updating for matrix

openrouteservice/pom.xml
```diff
@@ -403,6 +403,17 @@
         <version>2.9.2</version>
         <scope>compile</scope>
     </dependency>
+    <dependency>
```
rabidllama, line 406: Is there a reason why we need these? I just removed them from the pom and everything seemed to build/run ok

rabidllama (changes requested):
See comments

openrouteservice-api-tests/src/test/java/heigit/ors/v2/services/config/AppConfig.java
TimMcCauley: Yes, but then we have to make sure it's changed everywhere, would make sense to start here.

MichaelsJP: Everything related to the AppConfig.java reader was taken directly (with some exclusions) from the core. So if we want to change anything, we should start changing it in the core and I'll take it from there to stay compliant.

openrouteservice-api-tests/src/test/java/heigit/ors/v2/services/config/AppConfig.java
MichaelsJP: We could. But personally I would leave it in here. Only v2 package is using it so far and I don't think that this will change in the near future.

openrouteservice/pom.xml
MichaelsJP: When I remove it, the application fails to build

rabidllama (approved):

PR #357, opened 2018-11-16
Residential penalty may not reduce speed to lower than 5 (fixes #191 )

### LINKED ISSUE #191: Inconsistent responses from matrix API when route not found
I am receiving a duration of null when requesting the time-distance matrix between these two points:
```
$  http https://api.openrouteservice.org/matrix profile==driving-car api_key==my-api-key range==1 locations=='14.101061,53.29817|12.808634,52.927325' sources==0
```

Response:
```
{
  "durations": [
    [
      0,
      null
    ]
  ],
  "destinations": [
    {
      "location": [
        14.101061,
        53.29817
      ],
      "snapped_distance": 0.03
    },
    {
      "location": [
        12.808634,
        52.927325
      ],
      "snapped_distance": 0
    }
  ],
  "sources": [
    {
      "location": [
        14.101061,
        53.29817
      ],
      "snapped_distance": 0.03
    }
  ],
  "info": {
    "service": "matrix",
    "engine": {
      "version": "4.5.0",
      "build_date": "2018-03-26T14:15:39Z"
    },
    "attribution": "openrouteservice.org | OpenStreetMap contributors",
    "timestamp": 1528366746044,
    "osm_file_md5_hash": "783b5a0ed086d6cc3eb6f98df9d5d79b",
    "query": {
      "profile": "driving-car",
      "units": "m"
    }
  }
}
```

How should I interpret this? I didn't find anything about it in the API docs.

HendrikLeuschner:
### Pull Request Checklist

- [x] 7. If changes/additions are made to the app.config file, I have added these to the app.config.sample file along with a short description of what it is for, and documented this in the Pull Request (below).

Fixes #191 .

- Key functionality added: Speed may not be set to lower than 5 by residential penalty
- Reason for change: Old code resulted in 0 speed for edges with 5 original speed sometimes. Then routing does not work from/to those edges.

### DIFF
--- a/openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/flagencoders/ORSAbstractFlagEncoder.java
+++ b/openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/flagencoders/ORSAbstractFlagEncoder.java
@@ -29,7 +29,10 @@ public abstract class ORSAbstractFlagEncoder extends AbstractFlagEncoder {
 	}
 
 	double addResedentialPenalty(double baseSpeed, ReaderWay way) {
+		if (baseSpeed == 0)
+			return 0;
 		double speed = baseSpeed;
+
 		if(way.hasTag("highway","residential")) {
 			double estDist = way.getTag("estimated_distance", Double.MAX_VALUE);
 			// take into account number of nodes to get an average distance between nodes
@@ -38,9 +41,12 @@ public abstract class ORSAbstractFlagEncoder extends AbstractFlagEncoder {
 			if(interimNodes > 0) {
 				interimDistance = estDist/(interimNodes+1);
 			}
-			if(interimDistance < 100) {
+			if(interimDistance < 100 && speed > 10) {
 				speed = speed * 0.5;
 			}
+			//Don't go below 2.5 because it will be stored as 0 later
+			if(speed < 5)
+				speed = 5;
 		}
 
 		return speed;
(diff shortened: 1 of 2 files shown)


openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/flagencoders/ORSAbstractFlagEncoder.java
```diff
@@ -38,9 +41,12 @@ protected ORSAbstractFlagEncoder(int speedBits, double speedFactor, int maxTurnC
 			if(interimNodes > 0) {
 				interimDistance = estDist/(interimNodes+1);
 			}
-			if(interimDistance < 100) {
+			if(interimDistance < 100 && speed > 10) {
```
rabidllama: do we need to have this check? I ma just thinking that if we have a speed of say 7km/h, this will not get halved and so would stay at 7. But then a speed of 12km/h would be halved, meaning that it would then be seen as havng a slower speed than the 7km/h way. 

if we don't put this check in, then we would end up with 7/2 = 3.5km/h, but then in the nxt check (speed <5) it would be taken back up to 5 and so avoid the problems

rabidllama (changes requested):

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/flagencoders/ORSAbstractFlagEncoder.java
HendrikLeuschner: Done

rabidllama (approved):

PR #369, opened 2018-12-07
Bugfix #358/tolls hgv

### LINKED ISSUE #358: Tolls for HGV considered only when vehicle_type is set
At the moment in order to pick up hgv-specific tolls with `avoid: tollways` one has to additionally specify `vehicle_type` in options. Apart from the fact that this behavior us not documented, this is unpractical because currently the result doesn't depend on the particular value of `vehicle_type` anyway (even though it potentially could...).

Consider changing the logic such that results from `TollwayExtractor` depend on `profile` rather than `vehicle_type`.

aoles:
### Pull Request Checklist

- [x] 7. If changes/additions are made to the app.config file, I have added these to the app.config.sample file along with a short description of what it is for, and documented this in the Pull Request (below).

Fixes #358.

- Key functionality added:

Tolls are now determined based on the routing profile so in order for things to work there is no need to specify `vehicle_type` for HGV anymore.

Moreover, the `TollwaysGraphStorageBuilder` picks up any toll-specific tags, not just the first one encountered as it was the case in the previous approach. This allows to correctly deal with situations such as [here](https://www.openstreetmap.org/way/22765611) where next to a general tag `toll=no` a more detailed one `toll:hgv=yes` is present.

Because there are only a handful of toll tags relevant to the driving profiles, it was possible to reduce the total amount of space allocated per entry (edge) from 4 bytes to just one byte so a decrease of `ext_tolls` file size by a factor of 4 is expected.

- Reason for change: 

HGV tolls were resolved based on the presence of `vehicle_type` parameter rather than on the routing profile. This lead to tolls not being picked up unless `vehicle_type` was specified even though it's value was anyway not taken into account when extracting the tolls. On top of this combinations of toll tags were not parsed, just the first encountered tag.

Results for `driving-car` unlike for `driving-hgv` are in general not expected to change. For the latter around 1/3 off random routes on DACH are affected. This is because the toll data now being taken into account is more complete which is reflected either in additional entries in tolls extra info, or in different routes returned for `avoid_features=tollways`.

As an example consider driving hgv with `avoid_features=tollways` on the following [route](https://maps.openrouteservice.org/directions?n1=51.004353&n2=12.377365&n3=9&a=50.4217,12.638499,51.577923,11.895447&b=4a&c=0&k1=en-US&k2=km). The previous route (blue) changes (red) because now roads tagged with [`toll:N3`](https://overpass-turbo.eu/s/Emu) are being discarded.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Improved range and resolution of values encoding dimension/weight road restrictions in order to properly resolve them when corresponding hgv parameters are set (fixes issue #263)
 - Fixed empty BBox error if the route is located in the southern hemisphere (Issue #348)
 - Take into account access restrictions specific to hgv subprofiles (fixes issue #235)
+- Properly resolve all tolls, especially hgv-specific ones (fixes issue #358)
 ### Changed
 - Allowed access for cars and hgvs on access=destination roads (Issue #342)
 ### Deprecated
--- a/openrouteservice-api-tests/src/test/java/heigit/ors/services/routing/ResultTest.java
+++ b/openrouteservice-api-tests/src/test/java/heigit/ors/services/routing/ResultTest.java
@@ -787,10 +787,16 @@ http://localhost:8080/ors/routes?
 				.assertThat()
 				.body("any { it.key == 'routes' }", is(true))
 				.body("routes[0].containsKey('extras')", is(true))
-				.body("routes[0].extras.tollways.values.size()", is(1))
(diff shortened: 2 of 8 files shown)


rabidllama (approved):

rabidllama:
@aoles please let @TheGreatRefrigerator know and record in documentation about the changes to the tollways extended storage/extra info if it needs updating on the front end etc.

aoles:
> @aoles please let @TheGreatRefrigerator know and record in documentation about the changes to the tollways extended storage/extra info if it needs updating on the front end etc.

Thank you! There is no need to change anything on the front end because the tollway extra info is a binary yes/no information. After the update just additional route segments might be flagged as tollways.

PR #403, opened 2019-01-28
Feature #377 alternative routes

### LINKED ISSUE #377: Enable GraphHopper's alternative wayfinding methods
Let's evaluate the dev effort for

- round trip
- alternative routes

Ref. https://github.com/graphhopper/graphhopper/blob/master/docs/web/api-doc.md#flexible

Not sure who to assign here. @sfendrich? What do you think? Smth to look forward to in the new year?;)

takb:
### Pull Request Checklist

Fixes #377 .

- Key functionality added:
- routing API option "alternative_routes": n (int) added, uses GH alternative_route algorithm to return up to n routes for the query. Currently disables CH, and more than 2 waypoints are not supported. Parameters for the algorithm can be set with options "alternative_routes_weight_factor": (double) and  "alternative_routes_share_factor": (double).  

- Reason for change: 
This gets the algorithm to work basically, but a) the algorithm produces some funny results and b) CH support is underway (https://github.com/graphhopper/graphhopper/pull/1524), so we'll need to continue work on this feature in a while. The changes to RouteResultBuilder are also crying out for serious refactoring... So this is not done or anything, just stashing away interim results.

### DIFF
--- a/openrouteservice/src/main/java/heigit/ors/routing/RouteResultBuilder.java
+++ b/openrouteservice/src/main/java/heigit/ors/routing/RouteResultBuilder.java
@@ -57,11 +57,7 @@ public class RouteResultBuilder
     public RouteResult[] createRouteResults(List<GHResponse> routes, RoutingRequest request, List<RouteExtraInfo> extras) throws Exception {
         if (routes.isEmpty())
             return new RouteResult[]{new RouteResult(request.getExtraInfo())};
-
         if (routes.size() > 1) {
-            if (request.getSearchParameters().getAlternativeRoutes() > 1) {
-                throw new InternalServerException(RoutingErrorCodes.INVALID_PARAMETER_VALUE, "Alternative routes algorithm does not support more than two way points");
-            }
             return new RouteResult[]{createRouteResult(routes, request, extras)};
         }
         return createRouteResultSet(routes.get(0), request, extras);
--- a/openrouteservice/src/main/java/heigit/ors/routing/RouteResultBuilder.java
+++ b/openrouteservice/src/main/java/heigit/ors/routing/RouteResultBuilder.java
@@ -23,538 +23,221 @@ package heigit.ors.routing;
 import com.graphhopper.GHResponse;
 import com.graphhopper.PathWrapper;
 import com.graphhopper.util.*;
-import com.graphhopper.util.shapes.BBox;
 import com.vividsolutions.jts.geom.Coordinate;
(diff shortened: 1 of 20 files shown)


openrouteservice/src/main/java/heigit/ors/routing/RouteResultBuilder.java
```diff
@@ -54,297 +54,512 @@ public RouteResultBuilder()
...
+        boolean formatInstructions = request.getInstructionsFormat() == RouteInstructionsFormat.HTML;
+        boolean includeDetourFactor = request.hasAttribute("detourfactor");
+        boolean includeElev = request.getIncludeElevation();
+        DistanceUnit units = request.getUnits();
+        int unitDecimals = FormatUtility.getUnitDecimals(units);
+
+        RouteResult[] resultSet = new RouteResult[route.getAll().size()];
+
+        for (PathWrapper path : route.getAll()) {
+            RouteResult result = new RouteResult(request.getExtraInfo());
```
rabidllama, line 103: I would suggest that we refactor some of the code to reduce duplication of methods. For example, a lot of the code here is duplicated earlier, so it would be good if we could take this code and put it into a method that can be used for all places where this functionality is required. There are some differences relating to some aspects (e.g. loops), so it might need a bit of thought to get it working properly but it would certainly be a step in the right direction for cleaning up this class.

openrouteservice/src/main/java/heigit/ors/routing/RouteSearchParameters.java
```diff
@@ -65,6 +65,11 @@
     private int[] _avoidCountries = null;
     private BordersExtractor.Avoid _avoidBorders = BordersExtractor.Avoid.NONE;
 
+//  TAKB: parameters weight factor and share factor seem to be ignored by the algorithm, further testing required.
+    private int _alternativeRoutes = -1;
```
rabidllama: Would be better to name this _alternativeRouteCount to make it clearer what it is for (same for the getter and setter)

rabidllama (changes requested):
Mostly looks good, though I think we should do a small amount of refactoring in the RouteResultBuilder to reduce the duplication of the code that builds some of the route information.

There is a line of code in RoutingProfileManager (ln 463) that defines whether a route uses dynamic weights, which in turn is used to limit the length of routes requested. As the alternative routing algorithm does not used the optimised routing algorithms, we should add to that line a check for requesting alternative routes and if they are requested, then the distance check should also be carried out.

Issue #167, opened 2018-04-22
Lack of a manual/freehand route option


Martin868:
On all the routing engines I know that allow the user to draw a route, there is on option to bypass routing for a section of the route and manually link two points. This option is handy when the route need to follow a missing way on the map, or that there is a routing error in the map database. I do a lot of cycling route planning and this happens all the time. For example the route planners in Strava, RideWithGPS, Garmin, all offer this. In ORS this should be a type of via point.

nilsnolde:
This is a frontend issue, please register it here:
https://github.com/GIScience/openrouteservice-app/issues

TimMcCauley:
Frontend will consume this but it’s a feature that has to be implemented to the backend ultimately. 

The idea is: if a direct connection should be returned the backend simply doesn’t optimize between those two waypoints. We have the code for this in an older version (which was implemented), I might be able to find it.

dlichtistw:
I agree that this would be really useful for hiking route planning, too. I used it earlier, and am missing it now.

rabidllama:
@TimMcCauley Did you find where the old code was for this?

TimMcCauley:
@rabidllama sorry, I checked and I can't find it (most likely on my old laptop which was stolen 👎). The idea is quite straightforward, every waypoint gets an additional attribute optimize which is either `null` (startpoint), `true` (find an optimized path between it and the previous wp) or `false` which won't optimize but return the input coordinates (between it and the previous wp) as flight of the crow.

```
wp1 | optimize: null
wp2 | optimize: true
wp3 | optimize: false

paths:
wp1 - wp2 => optimized path
wp2 - wp3 => flight of the crow
```

rabidllama:
OK, I don't think this will be too major a thing to add, so lets aim for getting it in there for v4.7

rabidllama:
Moved to 4.7.2

MichaelsJP:
`"coordinates": [[8.688812, 49.414352],[8.688169, 49.413193],[8.689349, 49.413291],[8.688898, 49.410652], [8.687375, 49.409549]]`
![auswahl_035](https://user-images.githubusercontent.com/22392104/52954911-a13b4e80-338b-11e9-8b57-70938d1a70c8.png)
If you add `"skip_segments": [3]` it will add a direct connection between the third and fourth coordinate (third segment).
[...]

PR #416, opened 2019-02-20
Feature #167 freehand route

MichaelsJP:
### Pull Request Checklist

- [ ] 1. I have merged the latest version of the development branch into my feature branch and all conflicts have been resolved.

Fixes # .

- Key functionality added: The option to skip segments on the post route request (Feature request #167 ). 
- Reason for change: 

- to skip a segment the variable "skip_segments:[{integer}]" has to be added. Integers in the array refer to connection between two coordinates in the request and start counting from e.g. 1 for the segment between the first and second coordinate. The integers must always be > than zero and <= than the length of coordinates -1. There can be multiple segments skipped at the same time as long as the length of the skip_segments variable is <= length of the coordinates -1.

response
- the skipped segment will be represented as a direct line between the two coordinates that belong to the skipped segment. If instructions are turned on, the skipped segment will be represented in the instructions as an individual segment with its proper instructions. The duration to travel the skipped segment will be averaged from the other segments and applied to the length of the skipped segment. If no other segments are available the duration is zeroed. 

-


openrouteservice/src/main/java/heigit/ors/api/requests/routing/RouteRequestHandler.java
```diff
@@ -99,6 +99,25 @@ public  RoutingRequest convertRouteRequest(RouteRequest request) throws StatusCo
             }
         }
 
+        if (request.hasSkipSegments()) {
+            String skipSegmentsName = "skip_segments";
+            routingRequest.setSkipSegments(request.getSkipSegments());
+            for (int i = 0; i < request.getSkipSegments().size(); i++) {
```
rabidllama: This would be cleaner in a method that is called by the convertRouteRequest method

openrouteservice/src/main/java/heigit/ors/api/requests/routing/RouteRequest.java
```diff
@@ -227,6 +227,14 @@
     @JsonIgnore
     private boolean hasSimplifyGeometry = false;
 
+    @ApiModelProperty(name = "skip_segments", value = "Specifies the segments that should be skipped in the route calculation. " +
```
rabidllama: You should create a constant that contains the "skip_segments" text and then reference that

openrouteservice/src/main/java/heigit/ors/api/requests/routing/RouteRequestHandler.java
```diff
@@ -99,6 +99,25 @@ public  RoutingRequest convertRouteRequest(RouteRequest request) throws StatusCo
             }
         }
 
+        if (request.hasSkipSegments()) {
+            String skipSegmentsName = "skip_segments";
```
rabidllama: As mentioned above - there should be a constant within the API's RouteRequest object that defines the name of the skip segments property

openrouteservice/src/main/java/heigit/ors/api/requests/routing/RouteRequestHandler.java
rabidllama: a foreach loop would be easier to read here

openrouteservice/src/main/java/heigit/ors/routing/RouteResultBuilder.java
```diff
@@ -110,6 +110,10 @@ public RouteResult createRouteResult(List<GHResponse> routes, RoutingRequest req
 			}
 		}
 
+		if (request.getSkipSegments() != null && !request.getSkipSegments().isEmpty()) {
+			result.addWarning(new RouteWarning(3));
```
rabidllama: Use the RouteWarning.SKIPPED_SEGMENTS constant here, don't pass in a number itself

openrouteservice/src/main/java/heigit/ors/routing/RoutingProfile.java
```diff
@@ -761,16 +761,18 @@ else if (bearings[1] == null)
...
+                resp = mGraphHopper.directRoute(req);
+            } else {
+                mGraphHopper.setSimplifyResponse(geometrySimplify);
+                resp = mGraphHopper.route(req);
+            }
+            if (DebugUtility.isDebug() && !directedSegment) {
                 System.out.println("visited_nodes.average - " + resp.getHints().get("visited_nodes.average", ""));
             }
+            if (DebugUtility.isDebug() && directedSegment) {
+                System.out.println("skipped segment - " + resp.getHints().get("skipped_segment", ""));
```
rabidllama: Should use the LOGGER here rather than System.out.println

openrouteservice/src/main/java/heigit/ors/routing/RoutingRequest.java
```diff
@@ -1,202 +1,214 @@
...
-
-	public void setSuppressWarnings(boolean suppressWarnings) {
-		_suppressWarnings = suppressWarnings;
-	}
-
-	public boolean isValid() {
-		return !(_coordinates == null);
-	}
-}
+/*  This file is part of Openrouteservice.
```
rabidllama, line 1: why is the whole of this file being marked as changed?

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
```diff
@@ -233,6 +245,61 @@ protected void cleanUp() {
         super.cleanUp();
     }
 
+    public GHResponse directRoute(GHRequest request) {
```
rabidllama: should be named better to describe what the method actually does

openrouteservice/src/main/java/heigit/ors/api/responses/routing/GeoJSONRouteResponseObjects/GeoJSONSummary.java
```diff
@@ -59,4 +64,16 @@ public GeoJSONSummary(RouteResult result, List<JSONSegment> segments, Map extras
     public JSONSummary getSummary() {
         return new JSONSummary(this.distance, this.duration);
     }
+
+    @JsonProperty("warnings")
+    public List<Map> getWarnings() {
+        List<Map> warningsMap = new ArrayList<>();
+        for (RouteWarning warning: warnings) {
+            Map<Object, Object> warningMap = new HashMap<>();
```
rabidllama: The first item can be declared as a String - Map<String,Object> warningMap...

rabidllama (changes requested):
Mostly some small changes to make the code cleaner

openrouteservice/src/main/java/heigit/ors/routing/RoutingProfile.java
MichaelsJP: Should I also adjust the old println calls above?

rabidllama: can do...

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
MichaelsJP: I renamed it to constructAirLineRoute...

openrouteservice/src/main/java/heigit/ors/routing/RoutingRequest.java
MichaelsJP: For some reason someone set the windows line separator (CRLF) as the files default. I changed it back to the ORS unix default (LF).

rabidllama: We should make sure that we don't do these sort of things in a commit for a specific feature as it makes it lamost impossible to see the functionality that has been introduced or changed - I have to go through the file line by line to look for differences

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
rabidllama: still sounds a bit weird to me... I will update it to constructFreeHandRoute as the term air line is not used anywhere else and so it is a bit confusing

openrouteservice/src/main/java/heigit/ors/routing/RoutingRequest.java
MichaelsJP: Sure, I get that. I can overwrite the changes and add that to a unique commit?!

rabidllama: It's OK for this one as I can see where things changed as it is a smaller file, but if it was a bigger file then it wouldn't work

rabidllama (approved):

PR #423, opened 2019-02-27
Added missing /directions endpoint to old api

rabidllama:
Added missing endpoint for /directions (after removing the items from web.xml)


TimMcCauley (approved):
🌮

PR #510, opened 2019-05-08
Bugfix #325 make reachfactor for isochrones more realistic

### LINKED ISSUE #325: Isochrones reachability using too high maxspeeds
The reachability factor for isochrones is too low due to maxSpeeds being too high for every profile. I assume it has got do with with 

`https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/flagencoders/nextgen/NextGenBikeCommonFlagEncoder.java#L147`

Being used here:

`https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/heigit/ors/isochrones/Isochrone.java#L49`

Example request:

`http://localhost:8082/openrouteservice-4.7.0/isochrones?attributes=area|reachfactor&format=json&location_type=start&locations=8.684177,49.423034&profile=cycling-regular&range=100&range_type=time&units=m`

with 50 km/h (it is using the downhill speeds I think) this will compute a reachability factor of only *0.1* as the maximum reach area is

`(14m/s * 100s)squared * math.PI= 5.309.291 square meters`

and the area reached on the network is 

`629.313 square meters`

resulting in such a low reachability.

nilsnolde:
### Pull Request Checklist

Fixes #325  .

- Key functionality added: Reachability is now calculated using mean speeds defined for each profile.
- Reason for change: Before it was using the max speeds of each profile, which resulted in unrealistically low reachability factors.

### DIFF
--- a/openrouteservice/src/main/java/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
+++ b/openrouteservice/src/main/java/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
@@ -16,9 +16,7 @@ package heigit.ors.isochrones.builders.concaveballs;
 import com.carrotsearch.hppc.IntObjectMap;
 import com.carrotsearch.hppc.cursors.IntObjectCursor;
 import com.graphhopper.coll.GHIntObjectHashMap;
-import com.graphhopper.routing.util.AbstractFlagEncoder;
-import com.graphhopper.routing.util.FootFlagEncoder;
-import com.graphhopper.routing.util.HikeFlagEncoder;
+import com.graphhopper.routing.util.*;
 import com.graphhopper.storage.GraphHopperStorage;
 import com.graphhopper.storage.NodeAccess;
 import com.graphhopper.storage.SPTEntry;
@@ -34,11 +32,19 @@ import heigit.ors.isochrones.IsochroneSearchParameters;
 import heigit.ors.isochrones.builders.AbstractIsochroneMapBuilder;
 import heigit.ors.routing.RouteSearchContext;
 import heigit.ors.routing.graphhopper.extensions.AccessibilityMap;
+import heigit.ors.routing.graphhopper.extensions.flagencoders.CarFlagEncoder;
+import heigit.ors.routing.graphhopper.extensions.flagencoders.FootFlagEncoder;
+import heigit.ors.routing.graphhopper.extensions.flagencoders.HeavyVehicleFlagEncoder;
 import heigit.ors.routing.graphhopper.extensions.flagencoders.WheelchairFlagEncoder;
+import heigit.ors.routing.graphhopper.extensions.flagencoders.bike.ElectroBikeFlagEncoder;
+import heigit.ors.routing.graphhopper.extensions.flagencoders.bike.MountainBikeFlagEncoder;
+import heigit.ors.routing.graphhopper.extensions.flagencoders.bike.RegularBikeFlagEncoder;
+import heigit.ors.routing.graphhopper.extensions.flagencoders.bike.RoadBikeFlagEncoder;
 import heigit.ors.util.GeomUtility;
 import org.apache.log4j.Logger;
 import org.opensphere.geometry.algorithm.ConcaveHull;
 
+import javax.validation.constraints.Null;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.TreeSet;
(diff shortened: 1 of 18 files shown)


openrouteservice/src/main/java/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
```diff
@@ -524,4 +537,41 @@ private void copyConvexHullPoints(Polygon poly)
...
 	}
+
+
+    /**
+     * Determines mean speed to calculate the reachfactor property.
+     *
+     * @param encoder The relevant {@link FlagEncoder} instance.
+     * @return mean speed constant
+     */
+    private double getMeanSpeed(FlagEncoder encoder){
```
rabidllama: It would probably be better to move this into the ORSAbstractFlagEncoder object and then force each flagencoder to override it. Then inside the getMeanSpeed method you just need to check if it is an instance of ORSABstractFlagEncoder, and if so then call the method from there to get the speed:
```java
if (encoder instanceof ORSAbstractFlagEncoder) {
  meanSpeed = ((ORSAbstractFlagEncoder) encoder).getMeanSpeed();
}
```
That way, you don't have to add another if statement here everytime we add a new FlagEncoder.

rabidllama (changes requested):

openrouteservice/src/main/java/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
nilsnolde: ah nice! yes, that's what I was looking for! Will do tmrw

nilsnolde: hope that's better!

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/flagencoders/ORSAbstractFlagEncoder.java
```diff
@@ -35,6 +36,10 @@ public void setConsiderElevation(boolean considerElevation) {
 		this.considerElevation = considerElevation;
 	}
 
+	public double getMeanSpeed() {
```
rabidllama: It would be better to define this as an abstract method that must be overridden by the sub classes as then it ensures that we will always have something that is meant to be returned. currently if someone was to use the ORSAbstractFlagEncoder class without overriding, the mean speed would be 0 (the default value for a double).

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/flagencoders/ORSAbstractFlagEncoder.java
```diff
@@ -22,6 +22,7 @@
 
 public abstract class ORSAbstractFlagEncoder extends AbstractFlagEncoder {
 	private final static Logger logger = LoggerFactory.getLogger(ORSAbstractFlagEncoder.class);
+	protected static double MEAN_SPEED;
```
rabidllama: If you declare the getMeanSpeed method as abstract, it doesn't have a body and so this variable isn't needed

openrouteservice/src/main/java/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
```diff
@@ -32,10 +32,9 @@
 import heigit.ors.isochrones.builders.AbstractIsochroneMapBuilder;
```
rabidllama, line 32: Do a quick cleanup of the imports as it seems some are not needed anymore

rabidllama (changes requested):

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/flagencoders/ORSAbstractFlagEncoder.java
nilsnolde: jup, thx for the instructions:)

openrouteservice/src/main/java/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
nilsnolde: done

rabidllama (approved):

PR #514, opened 2019-05-21
feature-#349 list traversal countries

### LINKED ISSUE #349: List countries a route traverses and add corresponding duration within

MichaelsJP:
### Pull Request Checklist

Fixes #349  .

- Key functionality added: Return the countries the route traverses under route extras.
- Reason for change: It's nice to know ;).

-

-

### DIFF
--- a/openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/reader/borders/CountryBordersReader.java
+++ b/openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/reader/borders/CountryBordersReader.java
@@ -13,6 +13,8 @@
  */
 package heigit.ors.routing.graphhopper.extensions.reader.borders;
 
+import com.graphhopper.util.PointList;
+import com.graphhopper.util.shapes.GHPoint3D;
 import com.vividsolutions.jts.geom.Coordinate;
 import com.vividsolutions.jts.geom.Geometry;
 import com.vividsolutions.jts.geom.Point;
@@ -379,6 +381,18 @@ public class CountryBordersReader {
      */
     public static int getCountryIdByISOCode(String code) {
         return currentInstance != null ? currentInstance.isoCodes.getOrDefault(code.toUpperCase(), 0) : 0;
+    }
+
+    public static CountryBordersPolygon[] getCountriesByPointList(PointList pointList) {
+        CountryBordersPolygon[] countries = new CountryBordersPolygon[0];
+        for (GHPoint3D point :
+                pointList) {
+            Coordinate coordinate = new Coordinate(point.lon, point.lat);
+            CountryBordersPolygon[] countries2 = currentInstance.getCandidateCountry(coordinate);
+            CountryBordersPolygon[] countries23 = currentInstance.getCandidateCountry(coordinate);
+
+        }
+        return countries;
     }
 
     /**
--- a/openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/reader/borders/CountryBordersReader.java
+++ b/openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/reader/borders/CountryBordersReader.java
@@ -13,11 +13,8 @@
  */
 package heigit.ors.routing.graphhopper.extensions.reader.borders;
 
-import com.graphhopper.util.PointList;
-import com.graphhopper.util.shapes.GHPoint3D;
 import com.vividsolutions.jts.geom.Coordinate;
 import com.vividsolutions.jts.geom.Geometry;
-import com.vividsolutions.jts.geom.Point;
 import heigit.ors.geojson.GeometryJSON;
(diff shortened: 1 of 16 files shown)


openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/reader/borders/CountryBordersReader.java
```diff
@@ -381,6 +383,18 @@ public static int getCountryIdByISOCode(String code) {
         return currentInstance != null ? currentInstance.isoCodes.getOrDefault(code.toUpperCase(), 0) : 0;
     }
 
+    public static CountryBordersPolygon[] getCountriesByPointList(PointList pointList) {
```
rabidllama: This method doesn't make much sense - you are creating three arrays of the polygons, but only ever returning the empty one...

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/storages/BordersGraphStorage.java
```diff
@@ -74,24 +76,28 @@ public void setEdgeValue(int edgeId, short borderType, short start, short end) {
...
+	 * @param edgeId Internal ID of the edge to get values for
+	 * @param prop   The property of the edge to get (TYPE - border type (0,1,2), START - the ID of the country
+	 *               the edge starts in, END - the ID of the country the edge ends in.
+	 * @return The value of the requested property
 	 */
 	public short getEdgeValue(int edgeId, Property prop) {
+		// TODO maybe implement a second function that accesses only the country data
 		long edgePointer = (long) edgeId * edgeEntryBytes;
-		short border = 0, start = 0, end = 0;
+		short border;
```
rabidllama: Why are we creating the variables and then instantiating them in later calls rather than doing it in one line?

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/storages/builders/BordersGraphStorageBuilder.java
```diff
@@ -139,35 +142,33 @@ public void processWay(ReaderWay way, Coordinate[] coords, HashMap<Integer, Hash
...
+             short start = 0;
+             short end = 0;
+             try {
+                 start = Short.parseShort(cbReader.getId(startVal));
+                 end = Short.parseShort(cbReader.getId(endVal));
+             } catch (Exception ignore) {
+             } finally {
+                 if (start != end) {
+                     type = (cbReader.isOpen(cbReader.getEngName(startVal), cbReader.getEngName(endVal))) ? (short) 2 : (short) 1;
+                     _storage.setEdgeValue(edge.getEdge(), type, start, end);
```
rabidllama: We don't need to call this once in each part of the if - it can be done just once after the if as all the if does is set the type value.

You can even remove the else statement by setting the type value to be the NO_BORDER when you create it at line 154.

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/storages/builders/BordersGraphStorageBuilder.java
```diff
@@ -257,6 +258,9 @@ public String getName() {
             // Replace the arraylist
             countries = temp;
         }
+//        if (countries.size() > 2){
```
rabidllama: Remove commented code

rabidllama (changes requested):
Have you also run this on a larger dataset that crosses real life borders to make sure that query times aren't heavily affected? (they shouldn't be but worth just checking)

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/reader/borders/CountryBordersReader.java
MichaelsJP: Yep. Good find. Was a leftover.

openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/storages/BordersGraphStorage.java
MichaelsJP: As far as I understood, primitives like short always derfault to zero if not initialized with something else. So no need to.

MichaelsJP: Absolutely right. I intentionally kept it the way it was done before.

MichaelsJP:
> Have you also run this on a larger dataset that crosses real life borders to make sure that query times aren't heavily affected? (they shouldn't be but worth just checking)

I'm currently running Ireland and northern Ireland.

MichaelsJP:
I compared the performance for the branches dev and feature and there are no differences. It's ready to be merged.

PR #515, opened 2019-05-23
Release 5.0.1

rabidllama:
Update master to 5.0.1


aoles:
@rabidllama worth to mention in the release notes that this version uses an updated version of GH with patched ALT, leads to differences in routes (now correct) especially for walking profiles.

TimMcCauley (approved):

PR #588, opened 2019-09-20
Link changelog issues

TheGreatRefrigerator:
added link_issues.py which will automatically link issue numbers
like '(#456456)' in the CHANGELOG.md to the actual github issue
so you are able to navigate easily to the discussion of the changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,126 +6,131 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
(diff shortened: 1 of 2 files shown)


nilsnolde (commented):
Very neat stuff amando! Even nice code:) Do that in 20 lines, Java!

However, I don't think we should add this script to the ORS repo. This belongs more to a precommit hook, which could be useful in other repos of GIScience. 

Or if we add it to the repo, pls create a folder `scripts`, don't just dump it into the root :carousel_horse: 

I'll let @rabidllama decide if he wants it in here.

PR #635, opened 2019-12-12
fix algo selection for alternative routes with CORE and no LM

takb:
Fixes issues with AlternativeRoutes in case the profile setup has Core enabled and no LM. Also adds a test for AR requests with restriction params.


aoles:
TY for the fix! For a given weighting which has both CH and ALT enabled I still observe fallback to A* with beeline rather than ALT when no route filters are enabled. Can you confirm?

aoles (changes requested):
For a given weighting which has both CH and ALT enabled for a route without any filters switched on I still observe `AlternativeRoute` to use A* with the beeline rather than the landmarks approximator. This doesn't happen when no CH is available just ALT.

sfendrich (approved):

takb:
The flag setting code in RoutingProfile is a real mess. Added explicit setting of the LM flag in order to not disrupt the logic for other routing methods.

PR #661, opened 2020-01-24
Added parameter to app.config and check for maximum allowed round trip distance

### LINKED ISSUE #658: Add distance restriction to round trip routing
Similar to the alternative routes algorithm, GH fallbacks to A-Star Bi-direction, and so there should be a restriction placed on the API that limits the maximum estimated distance of a round trip route to be a value defined in the app.config.

See issue #651 , though implementation should be simpler as it just needs a quick check in the RouteRequestHandler against the value in the app.config as done with other API parameters. A good default would be similar to previous maximum distance with dynamic weights (100km).

rabidllama:
### Pull Request Checklist

- [ ] 5. I have written JUnit tests for any new methods/cl asses and ensured that they pass.

Fixes #658  .

- Key functionality added: Added app.config parameter for maximum distance of round trip routes and added code to check against this and throw an exception if the value provided is larger
- Reason for change: The current round trip algorithm in GraphHopper does not use optimisations, and so longer routes would be much slower.

-

- Added parameter `"maximum_distance_round_trip_routes": xxx,`


takb (changes requested):
add the value also to docker/conf/app.config.sample

takb (approved):
looks good

Issue #664, opened 2020-01-28
System messages in response 


takb:
Implement a way to put a system message in the metadata section of response

takb:
### Proposed format for specification:  JSON object in app.config:

##### ors.system_message: Array of Objects

##### each object in system message array: 
```
{
    active: Boolean # to be able to keep inactive messages in the file
    text: String
    condition: Object #omittable, message sent with every request
}
```
##### condition object: 
```
{
    "condition_name": value,
}
```

all conditions will be AND-chained. For details see the [wiki page](https://github.com/GIScience/openrouteservice/wiki/Configuration-(app.config)#orssystem_message)

The messages should be processed in the order they are specified, and the first applicable message sent.


### Output should be added to the Metadata object of the response: 

#### Route / Matrix JSON: for V1, the metadata object is called info.
```json
{
    "metadata": {
        "attribution": "openrouteservice.org, OpenStreetMap contributors, tmc - BASt",
        "service": "routing",
        "timestamp": 1580397758118,
        "query": {
            "coordinates": [
                [
                    8.678613,
                    49.411721
                ],
                [
                    8.714733,
                    49.393267
                ],
                [
                    8.687782,
                    49.424597
                ]
            ],
            "profile": "cycling-regular",
            "preference": "fastest",
            "format": "json",
            "instructions": true,
            "elevation": true
        },
        "engine": {
            "version": "6.0.0",
            "build_date": "2020-01-30T15:21:16Z",
            "graph_date": "2020-01-29T10:57:22Z"
        },

        "system_message": "A system message would be added here"
    }, 



}
```

#### Route / Isochrone GEOJSON: for V1, the metadata object is called info.
```json
{
    "type": "FeatureCollection",



    "metadata": {
        "attribution": "openrouteservice.org, OpenStreetMap contributors, tmc - BASt",
        "service": "routing",
        "timestamp": 1580398024981,
        "query": {
            "coordinates": [
                [
                    8.675154,
                    49.407727
                ],
                [
                    8.675863,
                    49.407162
                ]
            ],
            "profile": "driving-car",
            "preference": "shortest",
            "format": "geojson"
        },
        "engine": {
            "version": "6.0.0",
            "build_date": "2020-01-30T15:21:16Z",
            "graph_date": "2020-01-29T10:56:58Z"
        },
        "system_message": "A system message would be added here"
    }
}
```

#### Route GPX:
```xml
[...]

PR #669, opened 2020-02-04
Feature #664 system messages

takb:
### Pull Request Checklist

- [X] 5. I have written JUnit tests for any new methods/classes and ensured that they pass.

Implements  #664 , see there for details

### DIFF
--- a/openrouteservice/src/main/java/org/heigit/ors/api/requests/matrix/MatrixRequest.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/api/requests/matrix/MatrixRequest.java
@@ -187,8 +187,9 @@ public class MatrixRequest {
     public Set<String> getMetricsStrings() {
         Set<String> ret = new HashSet<>();
         if (metrics != null) {
-        for (MatrixRequestEnums.Metrics metric : metrics)
-            ret.add(metric.name());
+            for (MatrixRequestEnums.Metrics metric : metrics) {
+                ret.add(metric.name());
+            }
         }
         return ret;
     }
--- a/openrouteservice/src/main/java/org/heigit/ors/api/requests/matrix/MatrixRequest.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/api/requests/matrix/MatrixRequest.java
@@ -19,15 +19,17 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
 import org.heigit.ors.api.requests.common.APIEnums;
 import org.heigit.ors.exceptions.ParameterValueException;
 import org.heigit.ors.matrix.MatrixErrorCodes;
 import org.heigit.ors.services.matrix.MatrixServiceSettings;
-import io.swagger.annotations.ApiModel;
-import io.swagger.annotations.ApiModelProperty;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 @ApiModel(value = "MatrixRequest", description = "The JSON body request sent to the matrix service which defines options and parameters regarding the matrix to generate.")
 @JsonInclude(JsonInclude.Include.NON_DEFAULT)
@@ -181,6 +183,16 @@ public class MatrixRequest {
     public MatrixRequestEnums.Metrics[] getMetrics() {
         return metrics;
     }
+
+    public Set<String> getMetricsStrings() {
+        Set<String> ret = new HashSet<>();
+        if (metrics != null) {
+        for (MatrixRequestEnums.Metrics metric : metrics)
+            ret.add(metric.name());
+        }
+        return ret;
+    }
+
 
     public void setMetrics(MatrixRequestEnums.Metrics[] metrics) {
         this.metrics = metrics;
--- a/openrouteservice/src/main/java/org/heigit/ors/api/responses/routing/gpx/GPXMetadataExtensions.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/api/responses/routing/gpx/GPXMetadataExtensions.java
@@ -1,0 +1,34 @@
+/*
+ * This file is part of Openrouteservice.
+ *
+ * Openrouteservice is free software; you can redistribute it and/or modify it under the terms of the
+ * GNU Lesser General Public License as published by the Free Software Foundation; either version 2.1
+ * of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this library;
+ * if not, see <https://www.gnu.org/licenses/>.
+ */
+
+package org.heigit.ors.api.responses.routing.gpx;
+
+import org.heigit.ors.api.requests.routing.RouteRequest;
+import org.heigit.ors.api.util.SystemMessage;
+
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement(name = "extensions")
+public class GPXMetadataExtensions {
+    @XmlElement(name = "system-message")
+    private String systemMessage;
+
+    public GPXMetadataExtensions() {}
+
+    public GPXMetadataExtensions(RouteRequest request) {
+        this.systemMessage = SystemMessage.getSystemMessage(request);
+    }
+}
--- a/openrouteservice/src/main/java/org/heigit/ors/api/util/SystemMessage.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/api/util/SystemMessage.java
@@ -124,6 +124,7 @@ public class SystemMessage {
             }
             catch (Exception e) {
                 // ignore otherwise incomplete messages entirely
+                LOGGER.warn(String.format("Invalid SystemMessage object in app.config %s.", message.toString().substring(18)));
             }
         }
         LOGGER.info(String.format("SystemMessage loaded %s messages.", messages.size()));
--- a/openrouteservice/src/main/java/org/heigit/ors/api/util/SystemMessage.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/api/util/SystemMessage.java
@@ -1,0 +1,293 @@
+package org.heigit.ors.api.util;
+
+import com.typesafe.config.ConfigObject;
+import org.apache.log4j.Logger;
+import org.heigit.ors.api.requests.isochrones.IsochronesRequest;
+import org.heigit.ors.api.requests.matrix.MatrixRequest;
+import org.heigit.ors.api.requests.routing.RouteRequest;
+import org.heigit.ors.config.AppConfig;
+import org.heigit.ors.matrix.MatrixMetricsType;
+import org.heigit.ors.routing.RoutingProfileType;
+import org.heigit.ors.routing.RoutingRequest;
+import org.heigit.ors.routing.WeightingMethod;
+
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.*;
+
+public class SystemMessage {
+    private static final Logger LOGGER = Logger.getLogger(SystemMessage.class.getName());
+    private static List<Message> messages;
+
+    private SystemMessage() {}
+
+    public static String getSystemMessage(Object requestObj) {
+        if (messages == null) {
+            loadMessages();
+        }
+        if (messages.isEmpty()) {
+            return "";
+        }
+        if (requestObj == null) {
+            requestObj = "";
+        }
+        RequestParams params = new RequestParams();
+        // V1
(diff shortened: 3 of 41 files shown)


openrouteservice/src/main/java/org/heigit/ors/api/requests/matrix/MatrixRequest.java
```diff
@@ -182,6 +184,16 @@ public boolean hasDestinations() {
         return metrics;
     }
 
+    public Set<String> getMetricsStrings() {
+        Set<String> ret = new HashSet<>();
+        if (metrics != null) {
+        for (MatrixRequestEnums.Metrics metric : metrics)
```
rabidllama: Indent the contents of the if statement

openrouteservice/src/main/java/org/heigit/ors/api/responses/routing/gpx/GPXMetadataExtensions.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.heigit.ors.api.requests.routing.RouteRequest;
+import org.heigit.ors.api.util.SystemMessage;
+
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement(name = "extensions")
+public class GPXMetadataExtensions {
+    @XmlElement(name = "system-message")
```
rabidllama, line 26: Is there a reason why for the json responses it is "system_message" and for gpx it is "System-message"?

openrouteservice/src/main/java/org/heigit/ors/api/util/SystemMessage.java
```diff
@@ -0,0 +1,293 @@
...
+        for (ConfigObject message : AppConfig.getGlobal().getObjectList("system_message")) {
+            try {
+                if (message.toConfig().getBoolean("active")) {
+                    List<Condition> conditions = new ArrayList<>();
+                    loadConditionsForMessage(message, conditions);
+                    messages.add(new Message(message.toConfig().getString("text"), conditions));
+                }
+            }
+            catch (Exception e) {
+                // ignore otherwise incomplete messages entirely
```
rabidllama, line 126: Should we log here that there was an error? I believe that this method only gets called once, so logging that there is a problem could help with debugging without affecting performance

openrouteservice/src/main/java/org/heigit/ors/matrix/MatrixMetricsType.java
```diff
@@ -57,17 +66,32 @@ public static String getMetricNameFromInt(int metric) {
...
+
+	public static Set<String> getMetricsNamesFromInt(int metric) {
+		switch (metric) {
+			case MatrixMetricsType.DURATION:
+				return new HashSet<>(Collections.singletonList(KEY_DURATION)) ;
+			case MatrixMetricsType.DISTANCE:
+				return new HashSet<>(Collections.singletonList(KEY_DISTANCE)) ;
+			case MatrixMetricsType.WEIGHT:
+				return new HashSet<>(Collections.singletonList(KEY_WEIGHT)) ;
+			case 3:
```
rabidllama: Might be better to add this number as a constant with the others to make this a bit cleaner

openrouteservice/src/test/java/org/heigit/ors/api/util/SystemMessageTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.heigit.ors.routing.RoutingRequest;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.ArrayList;
+
+public class SystemMessageTest {
+    @Test
+    public void testGetSystemMessage() throws ParameterValueException {
+        System.setProperty("ors_app_config", "../openrouteservice-api-tests/conf/app.config.test");
```
rabidllama: Should unit tests rely on data being loaded from an "external" file? I don't have enough experience with them to say one way or the other, but just want to make sure that this is the correct way of doing it.

I am having problems locally running this test though as part of the whole, as it seems that the system message static class is being started earlier (in MatrixResponseInfoTest which calls MatrixResponseInfo which in turn calls the SystemMessage class) without the app.config data, and so the message list is empty.

openrouteservice/src/main/java/org/heigit/ors/api/util/SystemMessage.java
```diff
@@ -0,0 +1,293 @@
+package org.heigit.ors.api.util;
```
rabidllama, line 1: In general, would it be better to keep the core handling of generating system messages outside of the api package? One of the main ideas of having the API in it's own package was that all the code did there was to convert information between an API request and the object needed by the ors core - anything more than this conversion should happen within the main code itself. So in my mind, it would be better to move the system message class out of the api package and into the heigit.ors.util package. Then it keeps the system messages a bit more independent from the api itself which may be better for if ors gets used as a library and then the API code could effectively be ignored.

rabidllama (changes requested):
See comments

openrouteservice/src/main/java/org/heigit/ors/api/responses/routing/gpx/GPXMetadataExtensions.java
takb: old habits, XML naming convention used to be lowercase with hyphens or camel case. But our GPX format is a mess anyway, there is a "distance-units" and a "build_date" element in the extension, and several elements in the global extensions belong in the metadata anyway if you ask me. In JSON objects I'd normally use camel case, because it's meant to be handled in javascript, but there are already underscores in use (more consistently, afaik), so I went for that. 
To clean this up we'd need API v3. Maybe we should collect things like that and do a planned cleanup update at some point?  
As for now, I'm really not used to underscores in XML anyway, probably because I've been in the web and javascript world too much. Hyphens feel like better style to me, though we should maybe discuss this and add to our overall coding style guide (which we don't have yet, yay).

openrouteservice/src/main/java/org/heigit/ors/api/util/SystemMessage.java
takb: I really don't think so, since the system messages are in my mind part of the API and independent from the core functionalities. If someone were to use ORS as lib and not use our API code, they'd also have no use for the system messages code.

openrouteservice/src/test/java/org/heigit/ors/api/util/SystemMessageTest.java
takb: I rather think that that is the appropriate way, instead many of our older unit tests depend on values set in a local app.config file that does not exist and you have to set up ORS_APP_CONFIG to get them to run. It would be better to fix all unit tests so that they run on the basis of the same file everywhere. 
That config.test referenced is afaik also the one used to run the service for the api tests, which is why I think the tests run fine on travis. The other option would be to have a app.config for tests in the openrouteservice project, but that would mean a third file to keep up to date every time you change something.

openrouteservice/src/main/java/org/heigit/ors/matrix/MatrixMetricsType.java
takb: This is messed up deprecated code that had the problem I don't know how long, Thought I'd keep unnecessary work minimal by just get it to work.

openrouteservice/src/main/java/org/heigit/ors/api/util/SystemMessage.java
takb: done

openrouteservice/src/main/java/org/heigit/ors/api/requests/matrix/MatrixRequest.java
takb: done

openrouteservice/src/main/java/org/heigit/ors/api/util/SystemMessage.java
sfendrich: As far as I understood, the messages are meant to inform the API-users about changes in the services such as the deprecation of the old API, of parameters or of profiles. All these examples are unrelated to the core, hence, the core would be the wrong place for these use cases. Do we have any use cases, where the core would like to inject such a message in the response?

Further, the class imports several classes from the API, which is inappropriate for a class in the core and would be a serious design flaw.

However, we can also question, whether the API would be the right place. All of the above examples are messages about the services we run and not necessarily about the functions provided by the API. So we could even argue to place the messages at an even more abstract level. However, with the current setup, I believe that putting them into the API would be easier.

openrouteservice/src/main/java/org/heigit/ors/matrix/MatrixMetricsType.java
sfendrich: Could `case 3` be replaced by `case MatrixMetricsType.DURATION | MatrixMetricsType.DISTANCE`?

openrouteservice/src/main/java/org/heigit/ors/services/matrix/requestprocessors/json/JsonMatrixRequestProcessor.java
```diff
@@ -13,24 +13,14 @@
...
-import javax.servlet.http.HttpServletResponse;
-
-import org.json.JSONArray;
-import org.json.JSONObject;
-
 import com.graphhopper.util.Helper;
 import com.vividsolutions.jts.geom.Coordinate;
-
+import org.heigit.ors.api.util.SystemMessage;
 import org.heigit.ors.common.StatusCode;
```
sfendrich, line 19: As said above, a core class should not import an API class.

openrouteservice/src/main/java/org/heigit/ors/services/routing/requestprocessors/json/JsonRoutingResponseWriter.java
```diff
@@ -17,6 +17,7 @@
 import com.graphhopper.util.Helper;
 import com.graphhopper.util.shapes.BBox;
 import com.vividsolutions.jts.geom.Coordinate;
+import org.heigit.ors.api.util.SystemMessage;
```
sfendrich, line 20: Again: core class should not import API classes.

openrouteservice/src/test/java/org/heigit/ors/api/util/SystemMessageTest.java
sfendrich: I believe there is no good solution at the moment. The whole setup with app.conf and testing is a mess. Cleaning this up would be a further large project. Using ORS_APP_CONFIG would make the tests depend on a volatile environment, therefore, I second @takb 's solution for now.

sfendrich (approved):
Please check whether the proposed solution for org/heigit/ors/matrix/MatrixMetricsType.java works.

openrouteservice/src/main/java/org/heigit/ors/api/util/SystemMessage.java
rabidllama: OK

openrouteservice/src/test/java/org/heigit/ors/api/util/SystemMessageTest.java
rabidllama: Did you check about the issue with that test failing? As it stands I still cannot build openrouteservice without removing the test

takb: I just fixed the old unit tests so that they rely on the same app.config.test from api-tests. Now all tests run as a whole without setting ORS_APP_CONFIG.

openrouteservice/src/main/java/org/heigit/ors/matrix/MatrixMetricsType.java
takb: sure. done.

openrouteservice/src/main/java/org/heigit/ors/services/matrix/requestprocessors/json/JsonMatrixRequestProcessor.java
takb: that's old deprecated v1 api response production that should have been in the api package only it didn't exist I guess. Wouldn't care fixing code that's deprecated anyway.

openrouteservice/src/main/java/org/heigit/ors/services/routing/requestprocessors/json/JsonRoutingResponseWriter.java
takb: ditto.

openrouteservice/src/main/java/org/heigit/ors/services/matrix/requestprocessors/json/JsonMatrixRequestProcessor.java
takb: ditto.

openrouteservice/src/main/java/org/heigit/ors/services/routing/requestprocessors/json/JsonRoutingResponseWriter.java
takb: ditto.

openrouteservice/src/test/java/org/heigit/ors/api/util/SystemMessageTest.java
takb: So, I added another small improvement to the unit tests by telling maven to copy the app.config.test from openrouteservice-api-tests. now the uni tests run from maven and directly through intelli without requiring ORS_APP_CONFIG. For API tests locally you can still use ORS_APP_CONFIG for your local setup. I even suspect we could remove the copying command form the Travis setup, but it runs as is.

rabidllama (approved):

PR #679, opened 2020-02-21
Fix way category assignment for ferry connections

### LINKED ISSUE #678: Improve way category assignment for ferry connections
Way categories used by the avoid features filter are resolved independently from way surface types included in extra info annotations. If the logic between these two places differs it might lead to situations where certain way types which are supposed to be filtered out are still being used for routing.

This is the case for ferry connections. Consider the following [example](https://maps.openrouteservice.org/directions?n1=12.533786&n2=124.29451&n3=12&a=12.585246,124.08577,12.491471,124.284425&b=0&l1=1&c=0&k1=en-US&k2=km). Note the **avoid ferries** option, yet almost the whole route is annotated as a **ferry**.

image

The problem in this case is that the [code](https://github.com/GIScience/openrouteservice/blob/e458485e59f7430142dabad9f4ac3744759d0be8/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/WayCategoryGraphStorageBuilder.java#L82-L95) of  `WayCategoryGraphStorageBuilder`
combines way **category tags** with way **access tags**, which might lead to ways not being correctly identified as ferries. Way access is resolved for a given profile by the respective flag encoder and only ways considered for routing are further processed by the storage builders.

aoles:
### Pull Request Checklist

Fixes #678 .

- Reason for change: in certain cases ferry connections were not properly categorized for the use with avoid features filter. As a result, they were still considered for routing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed invalid JSON and GeoJSON when including elevation ([#640](https://github.com/GIScience/openrouteservice/issues/640))
 - Added graph date for isochrones and matrix service and fixed the 0 output ([#648](https://github.com/GIScience/openrouteservice/issues/648))
 - Fixed memory issue at graph building ([#659](https://github.com/GIScience/openrouteservice/issues/659))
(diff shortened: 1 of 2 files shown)


takb (approved):
looks ok

PR #692, opened 2020-04-06
Pedestrian core alt only

HendrikLeuschner:
### Pull Request Checklist

- Key functionality added: Use only A* Beeline or ALT for roundtrips. Support for pedestrian profile using only Core-ALT.
- Reason for change: CH/CALT do not work properly with roundtrip. At the same time, the goal is to use solely Core-ALT as speedup. This PR enables using foot profile when building only Core-ALT. All routing cases are then covered by either Core-ALT or A* Beeline. Roundtrip produces different results for A* Beeline and ALT and Core-ALT and CH, respectively. That is because the usage is not proper for CALT/CH, and because the usage of the distance approximator for A* and ALT is wrong (in GH), leading to suboptimal and therefore ambiguous routes. That is why two test cases have to be covered, one if ALT is enabled for foot, one if it is only A* Beeline.

ALT
ALT route
A* Beeline
A* Beeline route
Core-ALT
Core-ALT route

All three should be the same. But Core-ALT does not work with the current implementation and ALT/A* produce two different results because of the non-admissible heuristic used in GH implementation.

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 - Querying shortest weighting can now use CH shortest preparation if available
 ### Changed
 - Refactor the algorithm selection process
+- Use ALT/A* Beeline for roundtrips. Enable Core-ALT-only for pedestrian profile.
 ### Deprecated
 
 ## [6.1.0] - 2020-03-06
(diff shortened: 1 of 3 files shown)


takb (approved):
looks good to me

HendrikLeuschner:
Let me do some more testing before you merge. All api tests pass, but we should test on longer routes.

PR #700, opened 2020-04-22
fix tag order dependency

### LINKED ISSUE #677: Fix/Refactor WaySurfaceTypeGraphStorageBuilder
The current code to interpret way tags to determine surface type produces in some cases different results for the same set of tags depending on the order these are read. 

As @HendrikLeuschner / @aoles were pointing out, the assignment of waytype ferry is also broken. 

The code is also hard to read and cycles through all tags unnecessarily and should be refactored.

takb:
### Pull Request Checklist

Fixes #677 .

- Reason for change:
In cases where a surface tag was present on ways but not recognized (had invalid values) the encoded value would not fall back to the derived value based on the waytype, which lead to indeterminate results depending on the order the tags were processed. Implemented a consistent fallback behavior.

### DIFF
--- a/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/WaySurfaceTypeGraphStorageBuilder.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/WaySurfaceTypeGraphStorageBuilder.java
@@ -27,6 +27,9 @@ import java.util.Iterator;
 import java.util.Map.Entry;
 
 public class WaySurfaceTypeGraphStorageBuilder extends AbstractGraphStorageBuilder {
+	public static final String TAG_HIGHWAY = "highway";
+	public static final String TAG_SURFACE = "surface";
+	public static final String TAG_ROUTE = "route";
 	private WaySurfaceTypeGraphStorage storage;
 	private final WaySurfaceDescription waySurfaceDesc = new WaySurfaceDescription();
 	protected final HashSet<String> ferries;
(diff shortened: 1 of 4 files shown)


sfendrich (approved):
Cool! This fix should also improve graph-building performance.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/WaySurfaceTypeGraphStorageBuilder.java
```diff
@@ -48,39 +51,26 @@ public GraphExtension init(GraphHopper graphhopper) throws Exception {
...
+		} else if (way.hasTag(TAG_HIGHWAY)) {
+			wayType = WayType.getFromString(way.getTag(TAG_HIGHWAY));
+		} else {
+			return;
+		}
+		waySurfaceDesc.setWayType(wayType);
 
-					if (waySurfaceDesc.getSurfaceType() == 0) {
-						if (wayType == WayType.ROAD ||  wayType == WayType.STATE_ROAD || wayType == WayType.STREET) {
-                            waySurfaceDesc.setSurfaceType((byte) SurfaceType.ASPHALT);
```
aoles, line 71: The assumed surface type `ASPHALT` is a more specific value than `PAVED` considered as the opposite of `UNPAVED`. I believe that if no `surface` tag is present  we should only stratify between the most general  complementing values `PAVED` and `UNPAVED`. This is backed up by to [OSM wiki](https://wiki.openstreetmap.org/wiki/Key:surface) stating: 
> For roads for motor vehicles there there is normally an assumption that the surface is `surface=paved` unless otherwise stated. Paved in OpenStreetMap is non-specific and may cover sealed, tarmac, asphalt, bitumen. `surface=unpaved` is treated as the opposite of paved.

aoles (changes requested):
Thank you, looks great! Please consider my comment regarding using `SurfaceType.PAVED` instead of `SurfaceType.ASPHALT`.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/WaySurfaceTypeGraphStorageBuilder.java
aoles: We will make the switch in a separate PR.

aoles (approved):

PR #709, opened 2020-05-06
read ele tag data for nodes if app.config flag set

takb:
### Pull Request Checklist

- Key functionality added:
Option in app.config ors.services.routing.elevation_preprocessed (true/false) added, if enabled ORSOSMReader reads elevation values from the 'ele' tags on nodes in the OSM file instead of using GHs elevation lookup. 
This is intended to be used in combination with the ors-preprocessor to reduce memory consumption during graph building. 

Also reduces processing time for the initial graph building, as a reference build times with germany, single profile build (car) with no graph preparations, elevation data for GH cached, local machine:
- original OSM file 589.679s
- preprocessed file without ele reading 378.853s
- preprocessed file with ele reading 339.448s.

### DIFF
--- a/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReader.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReader.java
@@ -23,6 +23,7 @@ import com.graphhopper.util.EdgeIteratorState;
 import com.graphhopper.util.shapes.GHPoint;
 import com.vividsolutions.jts.geom.Coordinate;
 import org.apache.log4j.Logger;
+import org.heigit.ors.config.AppConfig;
 import org.heigit.ors.routing.graphhopper.extensions.reader.osmfeatureprocessors.OSMFeatureFilter;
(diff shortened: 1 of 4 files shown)


rabidllama (approved):

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReader.java
```diff
@@ -412,5 +415,15 @@ protected void finishedReading() {
 		procCntx.finish();
 	}
 
-
+	@Override
+	protected double getElevation(ReaderNode node) {
+		if (processElevationFromPreprocessedData) {
+			double ele = node.getEle();
+			if (Double.isNaN(ele)) { // this should npt happen if preprocessing worked correctly
```
sfendrich: Typo in comment: npt -> not
Also if this should not happen, we should at least log a warning.

sfendrich (changes requested):

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReader.java
takb: fixed.

sfendrich (approved):

PR #720, opened 2020-05-12
[bugfix] Include the correct limit of routes in the matrix error response (fixes #719)

### LINKED ISSUE #719: Small bug with matrix maximum_routes error response
#### Here's what I did
Request more routes in the matrix service than `maximum_routes`, e.g. 51 points.

#### Here's what I got
I've got an error: `Request parameters exceed the server configuration limits. Only a total of 2601 routes are allowed.`

The error should contain the correct amount of maximum routes, not what the user has given. In my case 2500.

Fix the error message here:

https://github.com/GIScience/openrouteservice/blob/d1e158a96f1f8053c16f5d60c49772d8c4ccf784/openrouteservice/src/main/java/org/heigit/ors/api/requests/matrix/MatrixRequestHandler.java#L107

FinnStutzenstein:
Please note, if I have to do anything else/differently :D


sfendrich (approved):

sfendrich:
Thanks for the bugfix!

PR #723, opened 2020-05-16
Fixed underestimated suitability values in extra info

### LINKED ISSUE #722: Incorrect suitability values in extra info
Some suitability values returned in extra info end up underestimated. As [an example](https://maps.openrouteservice.org/directions?n1=50.019438&n2=9.068036&n3=15&a=50.014798,9.051791,50.018212,9.056969&b=1a&c=0&k1=en-US&k2=km) consider the following [road](https://www.openstreetmap.org/way/4006261) tagged with `class:bicycle=1`. This corresponds to `PriorityClass.PREFER` for which the normalized priority index value should be 8, however, the returned value is 7.

image

This happens because when the encoded priority values are scaled back to the 0-7 range they are cast to `int` by truncating the decimal part rather than by a proper rounding. It happens that for `PriorityClass.PREFER` the reconstructed value is `7.99` which in the current approach gets converted to `7` instead of `8`.

aoles:
### Pull Request Checklist

Fixes #722 .

Addresses the problem described in #722 by proper rounding with `Math.round()` rather than just plain type cast.


takb (approved):
ok

PR #740, opened 2020-06-03
Feature#480 maximum speed

### LINKED ISSUE #480: User adjustable speed capping
After removing `maximum_speed` we've received some feedback that in [certain scenarios](https://ask.openrouteservice.org/t/bug-maximum-speed-option-not-taken-in-account-for-isochrones-maps/305/7) this option has been actually useful.

It might be feasible to still support it even with CALT, but within a limited range such that it only affects high-speed roads, some of which are filtered out by `avoid=highways` anyway.

star-pirate:
### Pull Request Checklist

Fixes #480.

- Key functionality added:
- Reason for change: See #480 

- Added the option to compute routes using a maximum speed as a parameter/weighting.

- Addition of maximum_speed for the appropriate inclusion of edges in the core.


openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/routing/ResultTest.java
```diff
@@ -1222,6 +1222,29 @@ public void testOptimizedAndTurnRestrictions() {
...
 
+    @Test
+    public void testMaximumSpeedWeighting() {
+        JSONObject body = new JSONObject();
+        body.put("coordinates", constructCoords("8.63348,49.41766|8.6441,49.4672"));
+        body.put("preference", getParameter("preference"));
+        body.put("instructions", true);
+        body.put("user_speed",80);
+
+        // Test that the "right turn only" restriction at the junction is taken into account
```
sfendrich: Instead of a comment, rather use an auxiliary function with a meaningful name.
Also it is unclear how the code below actually tests what the comment says.

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/routing/ResultTest.java
```diff
@@ -1222,6 +1222,29 @@ public void testOptimizedAndTurnRestrictions() {
                 .statusCode(200);
     }
 
+    @Test
+    public void testMaximumSpeedWeighting() {
```
sfendrich: It is unclear how this test is related to MaximumSpeedWeighting. What does it actually test about the weighting?

openrouteservice/.gitignore
```diff
@@ -1,13 +1,30 @@
...
+!openrouteservice/src/main/resources/app.config.sample
+
+app.config.local
+cgiar_provider/
+graphs/
+target/
+.python-version
+nohup.out
+build/
+cgiar_cache/
```
sfendrich: How are the changes in this file related to the PR?

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequestHandler.java
```diff
@@ -504,4 +512,12 @@ private  boolean convertSetFlexibleMode(boolean useContractionHierarchies) throw
 
         return avoidCountryIds;
     }
+
+    private double convertUserSpeed(Double userSpeed) throws ParameterValueException{
```
sfendrich: The name "convert" does not reflect what the method actually does.

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequestHandler.java
```diff
@@ -504,4 +512,12 @@ private  boolean convertSetFlexibleMode(boolean useContractionHierarchies) throw
...
+        if(userSpeed < 80){
```
sfendrich: Why is such a lower bound necessary? If it is, then it should be configurable, not hard-coded.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
```diff
@@ -509,11 +509,11 @@ public RoutingAlgorithm createAlgo(Graph graph, AlgorithmOptions opts) {
         String algoStr = ASTAR_BI;
 
         if (ASTAR_BI.equals(algoStr)) {
-            CoreALT tmpAlgo = new CoreALT(graph, prepareWeighting, traversalMode);
+            CoreALT tmpAlgo = new CoreALT(graph,new PreparationWeighting(opts.getWeighting()), traversalMode);
```
sfendrich, line 512: How is this change related to the PR?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
```diff
@@ -509,11 +509,11 @@ public RoutingAlgorithm createAlgo(Graph graph, AlgorithmOptions opts) {
...
             tmpAlgo.setApproximation(RoutingAlgorithmFactorySimple.getApproximation(ASTAR_BI, opts, graph.getNodeAccess()));
             algo = tmpAlgo;
         } else if (DIJKSTRA_BI.equals(algoStr)) {
-            algo = new CoreDijkstra(graph, prepareWeighting, traversalMode);
+            algo = new CoreDijkstra(graph, new PreparationWeighting(opts.getWeighting()), traversalMode);
```
sfendrich, line 516: How is this change related to the PR?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
```diff
@@ -0,0 +1,50 @@
...
+import org.heigit.ors.routing.graphhopper.extensions.storages.HeavyVehicleAttributesGraphStorage;
+
+/**
+ * This class includes in the core all edges with speed more than the one set in the app.config file max_speed.
+ *
+ * @author Athanasios Kogios
+ */
+
+public class MaximumSpeedCoreEdgeFilter implements EdgeFilter {
+    private double maxSpeed =  ((AppConfig.getGlobal().getServiceParameter("routing.profiles.default_params","maximum_speed")) != null) ?   Double.parseDouble(AppConfig.getGlobal().getServiceParameter("routing.profiles.default_params","maximum_speed")) : 80; //Minimum speed of the core.
```
sfendrich: Please avoid long lines for better readability. 
The constant "80" should not be hard-coded.
The comment at the end of the line is not really useful.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
```diff
@@ -0,0 +1,50 @@
...
+    public final FlagEncoder flagEncoder;
+
+    public MaximumSpeedCoreEdgeFilter(FlagEncoder flagEncoder) {
+        this.flagEncoder = flagEncoder;
+        if (!flagEncoder.isRegistered())
+            throw new IllegalStateException("Make sure you add the FlagEncoder " + flagEncoder + " to an EncodingManager before using it elsewhere");
```
sfendrich: It looks like this exception should never been thrown, unless there is a programming error somewhere else in the application. Can't we capture such errors statically?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public boolean accept(EdgeIteratorState edge) {
+        if ( (edge.get(flagEncoder.getAverageSpeedEnc()) > maxSpeed) || (edge.getReverse(flagEncoder.getAverageSpeedEnc())) > maxSpeed ) { //If the max speed of the road is greater than that of the limit include it in the core.
```
sfendrich: Please reduce the line length. Is the comment really necessary?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
```diff
@@ -0,0 +1,116 @@
...
+import com.graphhopper.util.Parameters.Routing;
+
+/**
+ * This class creates the weighting for the routing according to the maximum speed set by user.
+ *
+ * @author Athanasios Kogios
+ */
+
+public class MaximumSpeedWeighting implements Weighting {
+    protected final static double SPEED_CONV = 3.6; //From km/h to m/s.
```
sfendrich: Use a more expressive name instead of a comment.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
```diff
@@ -0,0 +1,116 @@
...
+    private final double headingPenalty;
+    private final  double userMaxSpeed;
+    private final Weighting superWeighting;
+    private final DecimalEncodedValue avSpeedEnc;
+    private boolean calculateWeight;
+
+    public MaximumSpeedWeighting(FlagEncoder flagEncoder, HintsMap hintsMap, Weighting weighting) {
+        this.avSpeedEnc = flagEncoder.getAverageSpeedEnc();
+        this.superWeighting = weighting;
+        if(hintsMap.getDouble("user_speed",80) < 80  ){
```
sfendrich: Hard-coded constants again.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
```diff
@@ -0,0 +1,116 @@
...
+            throw new RuntimeException("User speed should be <= 80.");
+        }
+        this.headingPenalty = hintsMap.getDouble(Routing.HEADING_PENALTY, Routing.DEFAULT_HEADING_PENALTY);
+        this.userMaxSpeed = hintsMap.getDouble("user_speed",80);
```
sfendrich: Hard-coded constants again.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
```diff
@@ -0,0 +1,116 @@
...
+    private final  double userMaxSpeed;
+    private final Weighting superWeighting;
+    private final DecimalEncodedValue avSpeedEnc;
+    private boolean calculateWeight;
+
+    public MaximumSpeedWeighting(FlagEncoder flagEncoder, HintsMap hintsMap, Weighting weighting) {
+        this.avSpeedEnc = flagEncoder.getAverageSpeedEnc();
+        this.superWeighting = weighting;
+        if(hintsMap.getDouble("user_speed",80) < 80  ){
+            throw new RuntimeException("User speed should be <= 80.");
```
sfendrich: Hard-coded constants again.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
```diff
@@ -0,0 +1,116 @@
...
+        // add direction penalties at start/stop/via points
+        boolean unfavoredEdge = edge.get(EdgeIteratorState.UNFAVORED_EDGE);
+        if (unfavoredEdge)
+            time += headingPenalty;
+
+        return time;
+    }
+
+    /** This function is going to add the difference of: (edge speed - maximum_speed) in the calcMillis. */
+    private double calcMaximumSpeedMillis(double userMaxSpeed, double speed, EdgeIteratorState edge){
```
sfendrich, line 60: Does the method name match the method behavior?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
```diff
@@ -0,0 +1,116 @@
...
+
+        return time;
+    }
+
+    @Override
+    public double calcWeight(EdgeIteratorState edge, boolean reverse, int prevOrNextEdgeId) {
+        if (calculateWeight) {
+            double speed = reverse ? edge.get(avSpeedEnc) : edge.getReverse(avSpeedEnc);
+            if (speed > userMaxSpeed)
+                return calcMaximumSpeedWeight(userMaxSpeed, edge);// Use the maximum speed defined }
```
sfendrich: Does the comment add any information?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
```diff
@@ -0,0 +1,116 @@
...
+        return superWeighting.calcWeight(edge, reverse, prevOrNextEdgeId);
+    }
+
+    @Override
+    public long calcMillis(EdgeIteratorState edge, boolean reverse, int prevOrNextEdgeId){
+        double speed = reverse ? edge.get(avSpeedEnc) : edge.getReverse(avSpeedEnc);
+        if (speed > userMaxSpeed) {
+            return superWeighting.calcMillis(edge, reverse, prevOrNextEdgeId) + (long) calcMaximumSpeedMillis(userMaxSpeed, speed, edge);
+        } else {
+            return superWeighting.calcMillis(edge, reverse, prevOrNextEdgeId);// If the speed of the edge is zero or lower than the one defined by user we call the superWeighting calcWEight method.
```
sfendrich: Remove the useless comment.

openrouteservice/src/main/resources/app.config.sample
```diff
@@ -68,6 +68,7 @@
             "maximum_distance_alternative_routes": 100000,
             "maximum_alternative_routes": 3,
             "maximum_distance_round_trip_routes": 100000,
+            "maximum_speed": 80,
```
sfendrich: Should this configuration parameter be used instead of the hard-coded values above?

sfendrich (changes requested):

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/routing/ResultTest.java
star-pirate: It uses a speed (80) and computes a route that we know will differ in distance from the one without the maximum speed.

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/routing/ResultTest.java
star-pirate: The comment was left from the turn-restrictions test when I copied it. I wrote a new one. Auxiliary function?

openrouteservice/.gitignore
star-pirate: Must have been left from an older commit when I was working on development branch.

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequestHandler.java
star-pirate: I used the naming from the previous functions for a consistency. But since my feature is only a double I did not nedd to convert it to sth. Do you have any suggestions for a name?

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequestHandler.java
star-pirate: The edges on the core are set to more than 80 by default. I will parse the value from the app.config instead of hard-coding it.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
star-pirate: The prepareWeighting variable is set in the constructor during preprocessing and it is final so it cannot be changed when using CALT. Since MaximumSpeedWeighting may or may not be used I need to allow for a dynamic change of the weighting using the code I wrote.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
star-pirate: The prepareWeighting variable is set in the constructor during preprocessing and it is final so it cannot be changed when using CALT. Since MaximumSpeedWeighting may or may not be used I need to allow for a dynamic change of the weighting using the code I wrote.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
star-pirate: Split it into multiple lines.
Its is 80 as a safety valve in case the value for the maximum speed is missing from the app.config or cannot be parsed for some reason. But I can erase it.
Changed the comment.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
star-pirate: I have copied this part from other Filters when I firstly wrote this code. I erased it now.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
star-pirate: No not necessary but maybe helpful. Should I erase it?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
star-pirate: Done.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
star-pirate: It returns the time needed for the route only if the speed of the edge is bigger than the speed set by the user so I guess yes.

openrouteservice/src/main/resources/app.config.sample
star-pirate: Yes I changed that. Thanks.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
+* Added a Maximum Speed Core Edge Filter to include in the core edges with speed more than a value parsed from app.config.
```
aoles: Each pull request typically corresponds to a single bullet point in the CHANGELOG. Apart from that, this file is meant for end users, so it should highlight the actual API feature rather than any technical details behind it. In fact one could use the speed limit with other dynamic algos such as ALT too. So I think sth along the lines would be more appropriate here:
> New `max_speed` parameter to the driving profiles of the directions API to specify a speed limit above a certain threshold set in the config file.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -17,7 +17,7 @@ Fixes # .
 
 ### Information about the changes
 - Key functionality added:
-- Reason for change:
+- Reason for change: C
```
aoles: I believe the edit to this file happened by mistake; please revert the change.

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/routing/ResultTest.java
aoles: I agree with @sfendrich, as we are not directly testing the weighting but rather an API feature. I would rename the test just to `testMaximumSpeed`.

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/routing/ResultTest.java
```diff
@@ -1222,6 +1222,29 @@ public void testOptimizedAndTurnRestrictions() {
...
+        JSONObject body = new JSONObject();
+        body.put("coordinates", constructCoords("8.63348,49.41766|8.6441,49.4672"));
+        body.put("preference", getParameter("preference"));
+        body.put("instructions", true);
```
aoles: The test queries should be stripped to the bare minimum and focus on the actually tested functionality. In this context enabling the instructions seems obsolete to me.

openrouteservice/.gitignore
aoles: Please consider reverting the changes to this file.

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequest.java
```diff
@@ -57,6 +57,8 @@
     public static final String PARAM_SIMPLIFY_GEOMETRY = "geometry_simplify";
     public static final String PARAM_SKIP_SEGMENTS = "skip_segments";
     public static final String PARAM_ALTERNATIVE_ROUTES = "alternative_routes";
+    public static final String PARAM_USER_SPEED = "user_speed";
```
aoles: I suggest renaming the API parameter to a more descriptive name `maximum_speed`. In fact, this was the name of the original parameter we have removed a while ago, and which we are reimplementing here (with the added lower limit).

Don't forget to update the names of methods and fields not only here, but across other classes as well.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
```diff
@@ -523,6 +532,17 @@ public void postProcessing() {
 			coreEdgeFilter.add(new WheelchairCoreEdgeFilter(gs));
 		}
 
+		/* Maximum Speed Filter */
```
aoles, line 540: Please refactor the code responsible for enabling `MaximumSpeedCoreEdgeFilter` such that repetitive code is reduced.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
aoles: I can confirm that this change is indeed necessary.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
aoles: I can confirm that this change is indeed necessary.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+public class MaximumSpeedCoreEdgeFilter implements EdgeFilter {
+    private double maxSpeed =  ((AppConfig.getGlobal().getServiceParameter("routing.profiles.default_params","maximum_speed")) != null)
+            ?   Double.parseDouble(AppConfig.getGlobal().getServiceParameter("routing.profiles.default_params","maximum_speed"))
+            : 80; //If there is a maximum_speed value in the app.config we use that. If not we set a default of 80.
+    public final FlagEncoder flagEncoder;
+
+    public MaximumSpeedCoreEdgeFilter(FlagEncoder flagEncoder) {
+        this.flagEncoder = flagEncoder;
```
aoles: Please refactor to store the average speed encoder rather than the `FlagEncoder`, similarly as in `MaximumSpeedWeighting`.

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequestHandler.java
```diff
@@ -504,4 +512,13 @@ private  boolean convertSetFlexibleMode(boolean useContractionHierarchies) throw
 
         return avoidCountryIds;
     }
+
+    private double convertUserSpeed(Double userSpeed) throws ParameterValueException {
+        double maxSpeed = ((AppConfig.getGlobal().getServiceParameter("routing.profiles.default_params", "maximum_speed")) != null) ? Double.parseDouble(AppConfig.getGlobal().getServiceParameter("routing.profiles.default_params", "maximum_speed")) : 80; //Minimum speed of the core.
+        if (userSpeed < maxSpeed) {
+            throw new ParameterValueException(RoutingErrorCodes.INVALID_PARAMETER_VALUE, RouteRequest.PARAM_USER_SPEED);
```
aoles: Could we return an error message containing the value set by the user together with an information such as "The maximum speed must not be lower than 80km/h" where the value should reflect the actually configured limit?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSWeightingFactory.java
```diff
@@ -24,6 +24,7 @@
 import com.graphhopper.storage.index.LocationIndex;
 import com.graphhopper.util.Helper;
 import com.graphhopper.util.PMap;
+import org.heigit.ors.routing.graphhopper.extensions.weighting.MaximumSpeedWeighting;
```
aoles: Necessary?

openrouteservice/src/main/resources/app.config.sample
aoles: The name `maximum_speed` might not entirely reflect the purpose of the parameter. In fact, it might falsely indicate a globally enabled speed limit. I would rename it to sth more descriptive such as `maximum_speed_lower_bound`.

aoles (changes requested):
Thank you @star-pirate , please see the individual comments.

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/routing/ResultTest.java
sfendrich: Auxiliary function = helper function. It is cleaner to encapsulate a code section as a new function with a meaningful name than to use a comment, because the code then explains itself.

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/routing/ResultTest.java
```diff
@@ -1222,6 +1222,29 @@ public void testOptimizedAndTurnRestrictions() {
...
+        body.put("user_speed",80);
+
+        //Test that the distance of the computed route.
```
sfendrich: Instead of adding this comment, this could be used to name the test.

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequestHandler.java
sfendrich: What about validateUserSpeed?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
sfendrich: Then the doc-comment should be corrected to reflect this behavior.

sfendrich (changes requested):
See the open conversations

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
+* New `maximum_speed` parameter to the driving profiles of the directions API to specify speed limit above a certain threshold set in the config file.
```
aoles: Now when I read it again my original wording sounds a bit clunky to me.  Maybe change
>to specify speed limit above a certain

to
>for specifying a speed limit above the

Or come up with something better :)

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
```diff
@@ -0,0 +1,120 @@
...
+
+    public MaximumSpeedWeighting(FlagEncoder flagEncoder, HintsMap hintsMap, Weighting weighting) {
+        this.avSpeedEnc = flagEncoder.getAverageSpeedEnc();
+        this.superWeighting = weighting;
+        if(hintsMap.getDouble("user_speed",maxSpeed) < maxSpeed  ){
+            throw new RuntimeException("User speed should be <=" + maxSpeed);
+        }
+        this.headingPenalty = hintsMap.getDouble(Routing.HEADING_PENALTY, Routing.DEFAULT_HEADING_PENALTY);
+        this.userMaxSpeed = hintsMap.getDouble("user_speed",maxSpeed);
+        this.calculateWeight = (hintsMap.getWeighting() == "fastest");
```
aoles: I think it would be better to set the flag based on `superWeighting.getName()` rather than `hintsMap.getWeighting()`.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
```diff
@@ -523,6 +532,17 @@ public void postProcessing() {
...
+		if (routingProfileCategory !=0 & (encodingManager.hasEncoder("heavyvehicle") || encodingManager.hasEncoder("car-ors")) ) {
```
aoles: Why not simply use `routingProfileCategory == RoutingProfileCategory.DRIVING` as conditional?

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
```diff
@@ -887,6 +887,10 @@ else if (bearings[1] == null)
                 req.getHints().put("alternative_route.max_share_factor", searchParams.getAlternativeRoutesShareFactor());
             }
 
+            if(searchParams.hasMaximumSpeed()){
+                req.getHints().put("user_speed", searchParams.getMaximumSpeed());
```
aoles: Let's be consistent in naming and use the label `maximum_speed` rather than `user_speed` in the hints map as well.

sfendrich (approved):

sfendrich:
Thanks for your contribution!

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingRequest.java
```diff
@@ -211,6 +212,10 @@ public void setIncludeCountryInfo(boolean includeCountryInfo) {
 		this.includeCountryInfo = includeCountryInfo;
 	}
 
+	public void setMaximumSpeed(double maximumSpeed){this.maximumSpeed = maximumSpeed;}
```
aoles: Could you improve code formatting of the one-line methods and split it across lines such as:
```
public void setMaximumSpeed(double maximumSpeed) {
    this.maximumSpeed = maximumSpeed;
}

**Review by aoles — CHANGES_REQUESTED** [review]
Thanks for addressing my previous comments.

I've added some further points for your consideration.

Cheers!

File: openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java (hunk 30 above)
Comment (star-pirate) [reply openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java]: We talked with Hendrik and decided to do it only for these vehicles. but I can do it in general.

File: openrouteservice/src/main/java/org/heigit/ors/routing/RoutingRequest.java (hunk 32 above)
Comment (star-pirate) [reply openrouteservice/src/main/java/org/heigit/ors/routing/RoutingRequest.java]: Since a lot of other functions had one-line methods I followed this format. Should I change others as well?

File: openrouteservice/src/main/java/org/heigit/ors/routing/RoutingRequest.java [hunk 33]
```diff
@@ -211,6 +212,12 @@ public void setIncludeCountryInfo(boolean includeCountryInfo) {
 		this.includeCountryInfo = includeCountryInfo;
 	}
 
+	public void setMaximumSpeed(double maximumSpeed){
+		this.maximumSpeed = maximumSpeed;
+	}
+
+	public double getMaximumSpeed(){return maximumSpeed;}
```
aoles: Would you mind fixing the formatting if the other method as well?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
aoles: I undestand. My point was that it seems that the result of the more obscure expression `routingProfileCategory !=0 & (encodingManager.hasEncoder("heavyvehicle") || encodingManager.hasEncoder("car-ors"))` is actually the same as of `routingProfileCategory == RoutingProfileCategory.DRIVING`. The latter one is more compact and easier to comprehend, thus preferred.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
```diff
@@ -0,0 +1,49 @@
...
+import org.heigit.ors.config.AppConfig;
+
+/**
+ * This class includes in the core all edges with speed more than the one set in the app.config file max_speed.
+ *
+ * @author Athanasios Kogios
+ */
+
+public class MaximumSpeedCoreEdgeFilter implements EdgeFilter {
+    private double maxSpeed =  ((AppConfig.getGlobal().getServiceParameter("routing.profiles.default_params","maximum_speed_lower_bound")) != null)
```
aoles: The handling of `maximum_speed_lower_bound` value from the config file still needs some refactoring. Currently there are 3 different places over the code where this value is parsed and if missing the hard-coded value of 80 is used. The default value, even if hard-coded, should appear only once in the whole codebase.

I suggest to use `RouteProfileConfiguration` class to store the value of `maximumSpeedLowerBound`.  Have a look at the implementation of the `maximum_distance` parameter, it will be probably quite similar.

This change might require, for example, that `maxSpeed` value is resolved upstream in `ORSGraphHopper` from `CmdArgs` set in `RoutingProfile`,  and passed to `MaximumSpeedCoreEdgeFilter` as a constructor argument.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
```diff
@@ -0,0 +1,120 @@
...
+ */
+
+public class MaximumSpeedWeighting implements Weighting {
+    protected final static double SPEED_UNIT_CONVERTER = 3.6; //From km/h to m/s.
+    private final double headingPenalty;
+    private final  double userMaxSpeed;
+    private final Weighting superWeighting;
+    private final DecimalEncodedValue avSpeedEnc;
+    private boolean calculateWeight;
+    private double maxSpeed =  ((AppConfig.getGlobal().getServiceParameter("routing.profiles.default_params","maximum_speed_lower_bound")) != null)
```
aoles: As mentioned before, use `RouteProfileConfiguration` as a canonical place to store the dafault. `maxSpeed` could probably be passed via `HintsMap`.

Actually, validation of `maximum_speed` could happen upstream in `ORSGraphHopper` where `MaximumSpeedWeighting` is created, so there is no need for `maxSpeed` parameter here.

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequestHandler.java
```diff
@@ -504,4 +512,15 @@ private  boolean convertSetFlexibleMode(boolean useContractionHierarchies) throw
 
         return avoidCountryIds;
     }
+
+    private double validateMaximumSpeed(Double userSpeed) throws ParameterValueException {
+        double maximumSpeed = ((AppConfig.getGlobal().getServiceParameter("routing.profiles.default_params", "maximum_speed_lower_bound")) != null)
```
aoles: See the comment in `MaximumSpeedCoreEdgeFilter`.

aoles (changes requested):
Thanks a lot for taking your time to improving the code, we are almost there! 

First of all please make sure the API tests pass. Currently they fail, most probably because of  places where still the old parameter name `user_speed` is used. Please search for any leftover occurrences and prune them.

Apart from this, I believe the handling of the `maximum_speed_lower_bound` value set through the config file could still be improved - see the inline comments.

Cheers!

PS: Please have a look at any previously unresolved conversations as well, thanks!

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequestHandler.java
star-pirate: How can I pass it to `RouteRequestHandler` from the `RouteProfileConfiguration` ?

aoles: That's a fair point and I was wondering about it too. To me it it looks as a general design flaw of ORS. If this is the case it might be that the validation of the value provided in the query can be done only later in `ORSGraphHopper`.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
star-pirate: Which `HintsMap` exactly?

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequestHandler.java
aoles: BTW: apart from verifying whether the `maximum_speed` parameter value is not below `maximum_speed_lower_bound` we need as check whether the parameter is compatible with the given profile. Unless the profile is car or hgv providing `maximum_speed` should raise an error. I don't think this is the case now.

star-pirate: > BTW: apart from verifying whether the `maximum_speed` parameter value is not below `maximum_speed_lower_bound` we need as check whether the parameter is compatible with the given profile. Unless the profile is car or hgv providing `maximum_speed` should raise an error. I don't think this is the case now.

Yes I can fix it.So I should move the verification from `RouteRequestHandler` to `ORSGraphHopper` ?

aoles: >So I should move the verification from RouteRequestHandler to ORSGraphHopper 

Actually, I was wrong about `ORSGraphHopper`, sorry for that. I think the verification should rather happen in `RoutingProfileManager` which has access to `RouteProfileConfiguration`.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
aoles: Change of plans. After looking into this further I think `maximum_speed` validation should be handled even before in `RoutingProfileManager`. It should be safe to assume that once `MaximumSpeedWeighting` is created the value has been already verified so no need to check again.

star-pirate: So Should I wait for further info?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
aoles: Update: please use `GraphProcessContext` rather than `CmdArgs` for passing the `maximum_speed_lower_bound`.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
aoles: I don't have any more insights for now. My latest comments reflect how I would proceed with the implementation, however, feel free to look around yourself. Maybe there is an even better and cleaner way of doing this?

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingRequest.java
aoles: Yes, please, but only your new methods relevant for this PR.

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequestHandler.java
star-pirate: Do you know maybe in which function should I add it? Are all verifications done somewhere?

aoles: You could add it for example [here](https://github.com/GIScience/openrouteservice/blob/b3eef42da7fe06c5965459a2ce501415adf46bff/openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfileManager.java#L578).

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/MaximumSpeedWeighting.java
aoles: Moving the validation of `maximum_speed` upstream to `RoutingProfileManager` has rendered both `maximumSpeedLowerBound` and `maxSpeed` obsolete. If the program execution arrives at this point it is already clear that `maximum_speed` is above the threshold, so no need of checking this again, right? Please remove both fields from the class.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
aoles: The introduction of `maximumSpeedLowerBound` has made `maxSpeed` redundant, so please remove the latter.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/MaximumSpeedCoreEdgeFilter.java
```diff
@@ -0,0 +1,51 @@
...
+    private final DecimalEncodedValue avSpeedEnc;
+
+    public MaximumSpeedCoreEdgeFilter(FlagEncoder flagEncoder, double maximumSpeedLowerBound) {
+        this.maximumSpeedLowerBound = maximumSpeedLowerBound;
+        this.avSpeedEnc = flagEncoder.getAverageSpeedEnc();
+    }
+
+    @Override
+    public boolean accept(EdgeIteratorState edge) {
+        if ( (edge.get(avSpeedEnc) > maxSpeed) || (edge.getReverse(avSpeedEnc)) > maxSpeed ) {
```
aoles: Use `maximumSpeedLowerBound` instead of `maxSpeed`.

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingRequest.java
```diff
@@ -207,10 +208,18 @@ public boolean getIncludeCountryInfo() {
 		return includeCountryInfo;
 	}
 
-	public void setIncludeCountryInfo(boolean includeCountryInfo) {
+	public void setIncludeCountryInfo(boolean includeCountryImnfo) {
```
aoles: Please be more careful when committing your changes. This is not the first time you are modifying some random stuff in the codebase.. In this case this actually introduces a bug!

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfileManager.java
```diff
@@ -576,6 +576,10 @@ public RoutingProfile getRouteProfile(RoutingRequest req, boolean oneToMany) thr
             }
         }
 
+        if(searchParams.getMaximumSpeed() !=0 & searchParams.getMaximumSpeed() < config.getMaximumSpeedLowerBound()){
```
aoles: Please use here the method `searchParams.hasMaximumSpeed()` and the short-circuit operator `&&` rather than `&`.

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfileManager.java
```diff
@@ -576,6 +576,10 @@ public RoutingProfile getRouteProfile(RoutingRequest req, boolean oneToMany) thr
...
+            throw new RuntimeException("The maximum speed must not be lower than " + config.getMaximumSpeedLowerBound() + " km/h.");
```
aoles: A`ParameterValueException` should be thrown here.

aoles (changes requested):
Please make sure that your updates are not breaking the code. In fact, currently the API test for maximum speed does not pass. I leave it to you to figure out what the problem is. 

Actually, test driven development is a good way of ensuring software requirements. I've added some additional API tests which cover the following checks:

-  that `maximum_speed` value below the configured threshold results in an appropriate error,
- `maximum_speed_lower_bound` setting can be profile-specific,
-  use with unsupported profile raises an error.

As long as you follow my comments from the review and fix the existing problems the first two tests  should pass without the need of implementing anything additional.

The last test where `maximum_speed` is provided to the cycling profile will probably require some additional logic. Take your time. Don't forget to add appropriate `CUSTOM_KEYS` indicating the profiles for which `maximum_speed` parameter is valid to the description in `@ApiModelProperty`. This is important from the perspective of our auto-generated interactive documentation.

Cheers,
Andrzej

star-pirate:
> Please make sure that your updates are not breaking the code. In fact, currently the API test for maximum speed does not pass. I leave it to you to figure out what the problem is.
> 
> Actually, test driven development is a good way of ensuring software requirements. I've added some additional API tests which cover the following checks:
> 
>     * that `maximum_speed` value below the configured threshold results in an appropriate error,
> 
>     * `maximum_speed_lower_bound` setting can be profile-specific,
> 
>     * use with unsupported profile raises an error.
> 
> 
> As long as you follow my comments from the review and fix the existing problems the first two tests should pass without the need of implementing anything additional.
> 
> The last test where `maximum_speed` is provided to the cycling profile will probably require some additional logic. Take your time. Don't forget to add appropriate `CUSTOM_KEYS` indicating the profiles for which `maximum_speed` parameter is valid to the description in `@ApiModelProperty`. This is important from the perspective of our auto-generated interactive documentation.
> 
> Cheers,
> Andrzej

I just checked my last commit and there the test is still passing. Did you change anything in your last commit?

So I should make sure that maximum speed is only called in `car-ors` and `driving-hgv` ?

aoles:
> I just checked my last commit and there the test is still passing. Did you change anything in your last commit?

I've just double checked that with you latest commit e996df7 `ResultTest.testMaximumSpeed` fails, so maybe you had some local changes which you forgot to commit?

> So I should make sure that maximum speed is only called in `car-ors` and `driving-hgv` ?

Yes, as discussed before the API parameter `maximum_speed` should be valid only for these two driving profiles.

Cheers,
Andrzej

star-pirate:
> > I just checked my last commit and there the test is still passing. Did you change anything in your last commit?
> 
> I've just double checked that with you latest commit [e996df7](https://github.com/GIScience/openrouteservice/commit/e996df71d54234f5bc4ad28f9fe8156a93325ac8) `ResultTest.testMaximumSpeed` fails, so maybe you had some local changes which you forgot to commit?
> 
> > So I should make sure that maximum speed is only called in `car-ors` and `driving-hgv` ?
> 
> Yes, as discussed before the API parameter `maximum_speed` should be valid only for these two driving profiles.
> 
> Cheers,
> Andrzej

Just updated. Everything should be fine now.

aoles (approved):
Looks good now, thanks for your contribution! I've fixed some minor outstanding glitches in my last commits.

Cheers,
Andrzej

PR #745, opened 2020-06-13
742 fix docker

### LINKED ISSUE #742: New docker containers always use the default app.config
It's very annoying: whenever a new container is created (e.g. when just changing some env var in `docker-compose.yml`) the new container reads the default `app.config` provided by the image. Regardless if the user actually has a valid config in `./ors-conf`. 

Not hard to remedy:

The problem is that `mvn package` is run in the image itself, which packages the default `app.config` into the WAR file. So currently there's no way to access the `app.config` Tomcat will use since it's inside the WAR.

The lines 
```
    cp -f /ors-core/openrouteservice/target/*.war /usr/local/tomcat/webapps/ors.war && \ 
```
need to be in `docker-entrypoint.sh`, not in the `Dockerfile`, with some logic so we don't package needlessly if nothing changed.

nilsnolde:
### Pull Request Checklist

Fixes #742 .

- Key functionality added: 
  - default `app.config.sample` from `./openrouteservice/src/main/resources/app.config.sample` is used to build the image
  - be aware the image changes the `app.config` to 
    - only build `car` instead of all profiles
    - `init_threads: 1` since > 1 can cause issues
    - changes all paths to docker image paths
  - uses the default `heidelberg.osm.gz`
- Reason for change: Before we had two `app.config`, one for the native Java app, one for docker. This PR removes that maintenance burden.


nilsnolde:
some git shit happened. note to self: def squash merge!

LittleBoxOfChicken:
Coming from [the ask.openrouteservice.org thread](https://ask.openrouteservice.org/t/own-hosted-openrouteservice-instance-not-working/1800/9?u=littleboxofchicken).

This does indeed fix the issue for me. 😃
Thanks a bunch, Nils.

nilsnolde:
Great, thanks for the quick feedback, Jari!

MichaelsJP (approved):

MichaelsJP:
Looking good. Thanks @nilsnolde

PR #757, opened 2020-06-28
Add configuration parameter to enable elevation smoothing

### LINKED ISSUE #725: Use of elevation data
There are the following potential issues with how the elevation data is being handled by ORS.

## Elevation is not considered for estimating edge length
 
Currently edge lengths are calculated based on (lon, lat) 2D spatial coordinates instead of doing this in 3D by taking elevation into account as well. GH's default 3D computations are deliberately [disabled in ORS](https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReader.java#L62) across all profiles.

This approach seems suboptimal to me. It typically leads to a small underestimation of route distance. For example, the [following route](https://maps.openrouteservice.org/directions?n1=49.408247&n2=8.719403&n3=15&a=49.408889,8.712083,49.403524,8.728139&b=0&c=0&k1=en-US&k2=km) with current 2D calculations yields a distance of **4467.5 vs. 4837.7 m** once 3D calculations are enabled.

image

Interestingly, the same route on [graphhopper.com/maps](https://graphhopper.com/maps/?point=49.408889%2C8.712083&point=49.403524%2C8.728139&locale=en-US&vehicle=bike&weighting=fastest&elevation=true&turn_costs=false&use_miles=false&layer=Omniscale) is **4.54 km** long.

image

This might be due to a different elevation source of quality (resolution) superior to our data. 

Once the elevation profiles are compared another striking difference becomes evident: there are a few sections of descent in our elevation profile which are not present in the GH's profile. In fact, in reality the road up to Königsstuhl has a more or less steady climb so these altitude drops are rather unexpected. This brings us to the next issue:

## No smoothing is performed on elevation data

According to GH:

> Elevation data is read from rectangular tiles. Especially when going along a cliff, valley, or pass, it can happen that a small part of the road contains incorrect elevation data. This is because the elevation data is coarse and sometimes contains errors.
This can lead to incorrect ascend, descend, and distance calculation of a route.

Once smoothing is enabled, the estimated distance becomes **4606 m**. But more importantly, now the elevation profile looks much more plausible and the previously reported ascent and descent values change from **363 m** and **43 m**, respectively, to **378 m** and **6 m**.

image

This can be verified by comparing against [Google](https://www.google.com/maps/dir/49.408889,8.712083/49.403524,+8.728139/@49.40233,8.7115,15z/data=!4m7!4m6!1m0!1m3!2m2!1d8.728139!2d49.403524!3e0). Note their ascent and descent values match pretty close to our numbers with elevation smoothing enabled.

image

aoles:
### Pull Request Checklist

Fixes #725.

- Key functionality added: GH's elevation smoothing.

- Reason for change: improve quality of elevation data returned to the user.

- The change doesn't have any effect on the generated routes other than differences in the height data.
See https://github.com/GIScience/openrouteservice/issues/725 for details.

**Optional** routing profile parameter `elevation_smoothing` for toggling elevation smoothing on/off.
The default is not to perform the smoothing (`false`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 ### Added
 - New `maximum_speed` parameter to the driving profiles of the directions API, for specifying a speed limit, above a certain threshold set in the config file.
 - Polish translation ([#690](https://github.com/GIScience/openrouteservice/issues/690))
+- Configuration parameter to enable elevation smoothing ([#725](https://github.com/GIScience/openrouteservice/issues/725))
 ### Fixed
 - Fixed fallback to dynamic routing methods if bearings parameter set ([#702](https://github.com/GIScience/openrouteservice/issues/702))
 - Enable elevation interpolation for bridges and tunnels ([#685](https://github.com/GIScience/openrouteservice/issues/685))
--- a/openrouteservice-api-tests/conf/app.config.test
+++ b/openrouteservice-api-tests/conf/app.config.test
@@ -48,6 +48,7 @@
             elevation_provider: multi,
             elevation_cache_clear: false,
             elevation_cache_path: "openrouteservice-api-tests/data/elevation",
+            elevation_smoothing: false,
             instructions: true,
             maximum_distance: 100000,
             maximum_distance_round_trip_routes: 100000,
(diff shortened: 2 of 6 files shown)


takb (approved):
looks ok

PR #760, opened 2020-07-01
Refactor/addition weighting

sfendrich:
### Pull Request Checklist

- Restructre AdditionWeighting
- Fix erroneous duration computation in AdditionWeighting.calcMillis

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Fixed fallback to dynamic routing methods if bearings parameter set ([#702](https://github.com/GIScience/openrouteservice/issues/702))
 - Enable elevation interpolation for bridges and tunnels ([#685](https://github.com/GIScience/openrouteservice/issues/685))
-- Fixed erroneous computation of duration in AdditionWeighting.calcMillis
+- Fixed erroneous duration computation of soft weightings such as green and quiet weightings
 ### Changed
 - Improve recommended weighting for cycling and walking profiles ([#665](https://github.com/GIScience/openrouteservice/issues/665))
 - Restructure AdditionWeighting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Fixed fallback to dynamic routing methods if bearings parameter set ([#702](https://github.com/GIScience/openrouteservice/issues/702))
 - Enable elevation interpolation for bridges and tunnels ([#685](https://github.com/GIScience/openrouteservice/issues/685))
+- Fixed erroneous computation of duration in AdditionWeighting.calcMillis
 ### Changed
 - Improve recommended weighting for cycling and walking profiles ([#665](https://github.com/GIScience/openrouteservice/issues/665))
+- Restructure AdditionWeighting
 ### Deprecated
 
 ## [6.1.1] - 2020-06-02
--- a/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/AdditionWeighting.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/AdditionWeighting.java
@@ -20,146 +20,21 @@ import com.graphhopper.util.EdgeIteratorState;
 
 public class AdditionWeighting extends AbstractWeighting {
 	private Weighting superWeighting;
-    private WeightCalc weightCalc;
+	private Weighting[] weightings;
 
     public AdditionWeighting(Weighting[] weightings, Weighting superWeighting, FlagEncoder encoder) {
         super(encoder);
         this.superWeighting = superWeighting;
-        
-        int count = weightings.length;
-        if (count == 1)
-           weightCalc = new OneWeightCalc(weightings);
-        else if (count == 2)
-            weightCalc = new TwoWeightCalc(weightings);
-        else if (count == 3)
-            weightCalc = new ThreeWeightCalc(weightings);
-        else if (count == 4)
-            weightCalc = new FourWeightCalc(weightings);
-        else if (count == 5)
-            weightCalc = new FiveWeightCalc(weightings);
+        this.weightings = weightings.clone();
     }
     
-    public interface WeightCalc  {
-		 double calcWeight(EdgeIteratorState edgeState, boolean reverse, int prevOrNextEdgeId);
-		 long calcMillis(EdgeIteratorState edgeState, boolean reverse, int prevOrNextEdgeId);
-    }
-    
-    public static class OneWeightCalc implements WeightCalc
-    {
-    	private Weighting weighting;
-    	
-    	public OneWeightCalc(Weighting[] weightings)
-    	{
-    		weighting = weightings[0];
-    	}
-    	
-    	public double calcWeight(EdgeIteratorState edgeState, boolean reverse, int prevOrNextEdgeId)
-    	{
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -11,8 +11,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Fixed fallback to dynamic routing methods if bearings parameter set ([#702](https://github.com/GIScience/openrouteservice/issues/702))
 - Enable elevation interpolation for bridges and tunnels ([#685](https://github.com/GIScience/openrouteservice/issues/685))
+- Fixed erroneous computation of duration in AdditionWeighting.calcMillis
```
aoles: Do you think we could be less technical in the changelog in favor of a more practical description highlighting the changes from the end-user perspective?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/AdditionWeighting.java
```diff
@@ -20,146 +20,21 @@
...
-           weightCalc = new OneWeightCalc(weightings);
-        else if (count == 2)
-            weightCalc = new TwoWeightCalc(weightings);
-        else if (count == 3)
-            weightCalc = new ThreeWeightCalc(weightings);
-        else if (count == 4)
-            weightCalc = new FourWeightCalc(weightings);
-        else if (count == 5)
-            weightCalc = new FiveWeightCalc(weightings);
+        this.weightings = weightings.clone();
```
aoles, line 28: Is cloning necessary here? It might introduce some performance penalty.

openrouteservice/src/test/java/org/heigit/ors/routing/graphhopper/extensions/weightings/AdditionWeightingTest.java
```diff
@@ -0,0 +1,33 @@
...
+    @Test
+    public void sumOfConstants () {
+        ConstantWeighting const1 = new ConstantWeighting(1,10);
+        ConstantWeighting const2 = new ConstantWeighting(2,20);
+        ConstantWeighting const3 = new ConstantWeighting(3,30);
+        ConstantWeighting factor = new ConstantWeighting(10, 100);
+
+        ConstantWeighting[] weightingsShort = {const1, const2, const3};
+        AdditionWeighting additionWeightingShort = new AdditionWeighting(weightingsShort,factor,flagEncoder);
+        assertEquals(60, additionWeightingShort.calcWeight(null, false, 0),0.0001);
```
aoles: How about testing for `calcMillis` as well?

openrouteservice/src/test/java/org/heigit/ors/routing/graphhopper/extensions/weightings/AdditionWeightingTest.java
```diff
@@ -0,0 +1,33 @@
...
+        encodingManager = EncodingManager.create(new ORSDefaultFlagEncoderFactory(), FlagEncoderNames.CAR_ORS, 4);
+        flagEncoder = (CarFlagEncoder)encodingManager.getEncoder(FlagEncoderNames.CAR_ORS);
+    }
+
+    @Test
+    public void sumOfConstants () {
+        ConstantWeighting const1 = new ConstantWeighting(1,10);
+        ConstantWeighting const2 = new ConstantWeighting(2,20);
+        ConstantWeighting const3 = new ConstantWeighting(3,30);
+        ConstantWeighting factor = new ConstantWeighting(10, 100);
```
aoles: I know it's just a name, yet it confused me at first that the super weighting is called factor in this context..

aoles (changes requested):
Good job, thanks for taking the time to clean up this mess! Please consider my inline comments.

Cheers!

CHANGELOG.md
sfendrich: Would "Fixed erroneous duration computation for soft weightings" be fine?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/AdditionWeighting.java
sfendrich: It is necessary to preserve the behavior. Otherwise, if the call-site would change the array after construction of the addition weighting, the latter would change under the hood, which was not the case in the previous implementation.

openrouteservice/src/test/java/org/heigit/ors/routing/graphhopper/extensions/weightings/AdditionWeightingTest.java
sfendrich: You are right. The name `superWeighting` didn't seem to make sense in the previous implementation. Now that we clarified that `calcMillis` was wrong and what its true purpose is, it makes sense again.

openrouteservice/src/test/java/org/heigit/ors/routing/graphhopper/extensions/weightings/AdditionWeightingTest.java
sfendrich: Yes, now that we understand what `calcMillis` should do, I can add a test.

sfendrich: I just noticed the useless suffix "Short" and will remove it.

CHANGELOG.md
aoles: It might be not clear what the soft weightings actually are unless we add sth like "such as green and quiet weightings".

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/AdditionWeighting.java
aoles: True. But do we actually need an _array_ in first place? Originally the weightings are bundled into a `List`, so maybe just skip the format conversion altogether and just pass the list. The weightings list in `ORSWeightingFactory.createWeighting()` is a local variable created just for constructing `AdditionWeighting` and not referenced elsewhere, so even though I agree about the potential threat of post-modification, I don't think this is of an issue here. What do you think?

openrouteservice/src/test/java/org/heigit/ors/routing/graphhopper/extensions/weightings/AdditionWeightingTest.java
aoles: Yeah, I was a little bit surprised by this name too :)

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/AdditionWeighting.java
sfendrich: In principle I agree, but I would prefer to do this in a second step because it has a broader scope and a bigger impact. My intention with this PR was to fix the bugs and clean up the biggest mess in a small, self-contained PR. All changes are local to `AdditionWeighting` and keep the interfaces stable. For changing interfaces we need more tests to refactor safely. IMO the whole model around the soft weightings looks unsound and would need to be redone anyways.

aoles: Sure, it's ultimately your call. I was suggesting this because `AdditionWeighting` is called only by `createWeighting` method so adjusting the interface didn't sound like a big deal to me, still a local change in a sense. While the mathematical model behind the soft weighting needs revising, the suggested modification is just a change to the implementation. Personally I would have done it in this PR in order to get this off the table and focus on the logic in any follow up PRs.

aoles (approved):
LGTM, Thanks again!

takb (approved):
Looks good.

PR #767, opened 2020-07-15
Release 6.2

takb:


openrouteservice/pom.xml
```diff
@@ -30,7 +30,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.heigit.ors</groupId>
     <artifactId>openrouteservice</artifactId>
-    <version>6.1.1</version>
+    <version>6.2</version>
```
rabidllama: Should we have 6.2 or 6.2.0 here? I'm not sure what the best notation is as in the past we have used both

takb: you're right we should stick to one... personally I'd prefer not writing out the zeros, but since we stuck to writing them out in the changelog I assume it's best to write it out everywhere.

rabidllama (approved):

rabidllama:
I don't think there is anything else that needs modifying for the version bump

PR #772, opened 2020-07-20
Allow configuration of app.config at docker run time

### LINKED ISSUE #742: New docker containers always use the default app.config

EWouters:
Fixes #742, #745.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -23,7 +23,9 @@ fi
 if [ -d "/usr/local/tomcat/webapps/ors" ]; then
 	cp -f /ors-conf/app.config $tomcat_appconfig
 else
-	cp -f $source_appconfig /ors-conf/app.config
+	if [ ! -f /ors-conf/app.config ]; then
+		cp -f $source_appconfig /ors-conf/app.config
+	fi
 	echo "### Package openrouteservice and deploy to Tomcat ###"
 	mvn -q -f /ors-core/openrouteservice/pom.xml package -DskipTests && \
 	cp -f /ors-core/openrouteservice/target/*.war /usr/local/tomcat/webapps/ors.war


rabidllama:
Just been having a look at this, and I don't think that it does what it's meant to. The bit of the `docker-entrypoint.sh` that has been modified is meant (in the original code) to copy the app.config that was created by the initial run of openrouteservice to the host so that it can be modifed in the host, and then when the container is restarted, this host file is copied to the ors folder in the tomcat webapps and then the changes take effect. What the modified code does is copy the config used in the original build to the `ors-conf` folder if there isn't one there. 

I think what is meant is that if someone provides a conf file from the host when the container is run for the first time (by passing it in during run with `-v $PWD/conf:/ors-conf`), the host file is used in the build process. For that, I believe rather than 

```
	if [ ! -f /ors-conf/app.config ]; then
		cp -f $source_appconfig /ors-conf/app.config
	fi
```
it should be
```
	if [ ! -f /ors-conf/app.config ]; then
		cp -f $source_appconfig /ors-conf/app.config
	else
               cp -f  /ors-conf/app.config $source_appconfig
        fi
```
which would then copy the container created app.config to the `ors-conf` if there is no file there, else it copies the file from `ors-conf` to the location of the config file used for the initial build.

I could be wrong though in terms of what you are aiming for here...

EWouters:
I'm aiming to use a custom app.config. The image you provide on docker hub does not allow that for some reason. Also if I build the image myself you have to change the code to get a custom app.config loaded. Seems like a bug.

rabidllama:
Yes I agree that it is something that needs fixing in the code, the question was as to whether the code that I suggested is the better solution as file `/ors-conf/app.config` itself doesn't do anything unless it is in the location defined by `source_appconfig` (`source_appconfig=/ors-core/openrouteservice/src/main/resources/app.config`).

The initial code you suggested only copies the existing file used to build to the `ors-conf` folder if there is nothing already there. If there is something there (your custom config file) then nothing happens with it and the default file (`source_appconfig`) is used. If you want to take into account the file that is in the ors-conf folder, then this needs to replace the file at `source_appconfig` which the "else" part of my suggested code does.

docker-entrypoint.sh
```diff
@@ -23,7 +23,9 @@ fi
 if [ -d "/usr/local/tomcat/webapps/ors" ]; then
 	cp -f /ors-conf/app.config $tomcat_appconfig
```
EWouters, line 24: This copies the users app.config to `/usr/local/tomcat/webapps/ors/WEB-INF/classes/app.config` if `/usr/local/tomcat/webapps/ors` exists and then ignores it

docker-entrypoint.sh
```diff
@@ -23,7 +23,9 @@ fi
...
 else
-	cp -f $source_appconfig /ors-conf/app.config
```
EWouters, line 26: This copies the default app.config and ignores the users app.config

EWouters (commented):
The logic in `docker-entrypoint.sh` does not make much sense to me.

Can a contributor explain the intended logic of working with a app.config file?

EWouters:
> 
> 
> Yes I agree that it is something that needs fixing in the code, the question was as to whether the code that I suggested is the better solution as file `/ors-conf/app.config` itself doesn't do anything unless it is in the location defined by `source_appconfig` (`source_appconfig=/ors-core/openrouteservice/src/main/resources/app.config`).
> 
> The initial code you suggested only copies the existing file used to build to the `ors-conf` folder if there is nothing already there. If there is something there (your custom config file) then nothing happens with it and the default file (`source_appconfig`) is used. If you want to take into account the file that is in the ors-conf folder, then this needs to replace the file at `source_appconfig` which the "else" part of my suggested code does.

I think there is more wrong than just the else part.

docker-entrypoint.sh
rabidllama: Correct, so here the config within the ors-conf folder is only copied to the tomcat webapp if openrouteservice has already been built and deployed (else the ors folder wont exist in webapps).If the ors folder doesnt exist (ors hasn't been built yet) then it would do the other option (see next comment)

docker-entrypoint.sh
rabidllama: OK, I think I see now some more of the logic after looking into the Dockerfile and docker-compose file. So, for the user to define their own app.config, they wouldn't place it into the folder on the host mapped to the internal ors-conf folder (or if they did, wouldn't name it as app.config) but need to tell Docker the path of the file using the APP_CONFIG arg (pointing to a file on the host within the build context). The file defined by this then overwrites the app.config.sample file inside the container, which is then renamed to app.config so that ors can use it. Then the paths within it modified to point to the correct folders within the container (this all happens in the Dockerfile). What the docker-entrypoint stuff does is to copy that file to  `/ors-conf/app.config` so that the user can see within the host (as long as it is mapped as a volume) what config is being used and make modifications to it. Then when the container is restarted this config file gets copied to overwrite the one in tomcat which is where openrouteservice is actually running from.

So yes, I think what your code is doing is correct in that it overwrites the custom file if it doesnt exist, but to get the custom config to be used when ors is first built, it needs to be passed as the `APP_CONFIG` arg in docker and not placed in the ors-conf folder mapped to the host.

rabidllama:
OK, so I think then my brain got a bit confused with where the custom config comes from - it needs to be passed to docker as the APP_CONFIG build arg. I managed to get this all working locally as expected using your code now i understand the process

rabidllama (approved):

docker-entrypoint.sh
EWouters: Yes, that seems to be the current logic. However, that does not match any use case I can think of, as the further lines in the dockerfile scrutinize the custom app.config. Also this makes the image published on docker hub quite useless.

EWouters:
After the user clones the repo and edits the app.config they have to build the image and then, if they map the `conf` folder can see their own config. This does not make much sense.

I propose to get rid of the map data and app.config in the dockerfile and make the user supply both of those upon docker run, at least on the first run. When they restart the container the existence of `/usr/local/tomcat/webapps/ors` can be used to skip the import.

This should make things simpler and allow the user to use the docker hub image.

rabidllama:
I believe the original reason for mapping the conf folder was that a number of users struggled with the concept of having file paths within the config that were for inside the container and not for paths on the host, so by changing the config used for input within the Docker file and then outputting it, users could see what paths were used. Also, it was getting a nightmare with maintaining different default config files within the codebase as there was one for running from source, one for the docker implementation, and one for the api tests.

The reasons also for having the config and osm data defined in the Dockerfile was that it allows users to run openrouteservice straight out of the box for testing without needing to obtain their own OSM data - in effect it was meant as a quick "this is how you can use it and set it up", and then providing individual data and configs would be the next "more advanced" step (in effect, the initial run of the docker image is to create the bare-bones testing environment)

I agree that things should be simple to run from the docker hub image, but I am not sure if forcing people to provide their own OSM data is the right way to go as then they need to at least have some basic concept of how to handle OSM data (most regions that can be downloaded from services such as GeoFabrik are pretty big and not suitable for a first time test) before they even get to using openrouteservice.

EWouters:
Ok, so I understand that you want the image to be able to run a demo out of the box, but also be usable as a base for custom app.configs.

In that case it would make sense to copy the app.config in the /ors-conf directory in the image to the $source_appconfig path if it exists.

This does not allow a user that does not supply an app.config to view the default app.config, but that use case is less important than the custom app.config use case in my opinion. This can easily be clarified in the documentation.

The logic is then:
```
# if Tomcat built before, copy the mounted app.config to the Tomcat webapp app.config, else copy it from the source
if [ -d "/usr/local/tomcat/webapps/ors" ]; then
	cp -f /ors-conf/app.config $tomcat_appconfig
else
	# if the user mounts an app.config file in /ors-conf, use the custom app.config
	if [ -f /ors-conf/app.config ]; then
		cp -f /ors-conf/app.config $source_appconfig
	fi
	echo "### Package openrouteservice and deploy to Tomcat ###"
	cp -f /ors-core/openrouteservice/target/*.war /usr/local/tomcat/webapps/ors.war
fi
```

I'll make a new PR (#784) for this fix.

PR #778, opened 2020-08-01
feat: improve road bike speeds

steventebrinke:
- Key functionality added: Improved road bike average speeds
- Reason for change: Road bike routing is very strange at the moment: a paved footway has speed 18, whereas a paved cycleway only has speed 8. This seems to be the wrong way around.

### Changes
- Speed limits are set only once (these were set twice, which made it harder to read, as some values were unused).
- Bad road surfaces decrease average speed: so an unclassified cobblestone road is slow due to the cobblestones (not fast because it is unclassified)
- Good road surfaces don't increase average speed: so a paved footway is slow because it's a footway (not fast because it is paved)

### Effect

When comparing the speeds, we see the following differences (before this PR -> after this PR):

|   | paved / asphalt / concrete | concrete:lanes / concrete:plates | cobblestone / cobblestone:flattened / paving_stones / paving_stones:30 / sett |
| - | -------------------------- | -------------------------------- | ----------------------------------------------------------------------------- |
| trunk |  | 20 -> 16 | 20 -> 10 |
| trunk_link |  | 20 -> 16 | 20 -> 10 |
| primary |  | 22 -> 16 | 22 -> 10 |
| primary_link |  | 22 -> 16 | 22 -> 10 |
| secondary |  | 24 -> 16 | 24 -> 10 |
| secondary_link |  | 24 -> 16 | 24 -> 10 |
| tertiary |  | 26 -> 16 | 26 -> 10 |
| tertiary_link |  | 26 -> 16 | 26 -> 10 |
| motorway |  | 18 -> 16 | 18 -> 10 |
| motorway_link |  | 18 -> 16 | 18 -> 10 |
| unclassified |  | 20 -> 16 | 20 -> 10 |
| road | 20 -> 12 | 20 -> 12 | 20 -> 10 |
| residential |  | 18 -> 16 | 18 -> 10 |
| service | 8 -> 12 | 8 -> 12 | 8 -> 10 |
| cycleway | 8 -> 18 | 8 -> 16 | 8 -> 10 |
| footway | 18 -> 6 | 16 -> 6 | 10 -> 6 |
| pedestrian | 18 -> 6 | 16 -> 6 | 10 -> 6 |
| path | 18 -> 8 | 16 -> 8 | 10 -> 8 |
| living_street | 18 -> 6 | 16 -> 6 | 10 -> 6 |
| steps | 18 -> 2 | 16 -> 2 | 10 -> 2 |
| track | 18 -> 2 | 16 -> 2 | 10 -> 2 |


sfendrich:
Thanks a lot for your PR! We will need some time to review it, in particular, because it removes code that was marked as required.

marq24:
Just as first comment to start a discussion - RoadBikeCycling profile is trying to avoid cycleways and footways by _purpose_ (and only use them if there are not alternatives). Specially in roundabout situations the osm cycleways is none optimal (even I am fully aware, that this routing is not StVO compliant - always take 20,- € with you). If this is not a option for you - use the default cycle profile.

marq24:
@sfendrich Could you please be so kind and let me know what was your key motivation to start this PR? Did you had unexpected routing results? Was your motivation to make the code easier to understand? Did you want to get closer "ride time" calculations?

And concerning your comment
> Bad road surfaces decrease average speed: so an unclassified cobblestone road is slow due to the cobblestones (not fast because it is unclassified)
My personal experience is, that surface have a massive effect for my speed (with the thin wheels) - much more then the road type/classification - I am not faster on a primary road then on a service was and vise versa.

> Good road surfaces don't increase average speed: so a paved footway is slow because it's a footway (not fast because it is paved)
Also this is IMHO the correct way - as road bike cyclist I try to avoid dedicated footways - no matter of their surface) - the current code give all all way types a massive "speed penalty" in order to try to avoid them (and stay on the streets)... 

I am totally open for discussion here - if you have example routings that end (in your opinion) in false turn by turn information - please share them.

For sure using higher speeds on cycleways (for road bikes) will end in routes like this one here:
https://route.emacberry.com/directions?n1=51.964534&n2=8.248345&n3=17&a=51.965245,8.244005,51.966388,8.251805&b=1a&c=0&k1=en-US&k2=km

and I am confident that a road bike cyclist would highly prefer this route here:
https://route.emacberry.com/directions?n1=51.964534&n2=8.248345&n3=17&a=51.965245,8.244005,51.966388,8.251805&b=1c&c=0&k1=en-US&k2=km

sfendrich:
Maybe @steventebrinke  can answer your questions, as it is his PR. Also I don't see where the cited comment comes from?

marq24:
> Maybe @steventebrinke can answer your questions, as it is his PR. Also I don't see where the cited comment comes from?

it was just too hot - or I am just too old - sorry - of course my questions goes to @steventebrinke and not you...

steventebrinke:
> Did you had unexpected routing results?

Yes. When routing for road bikes, the routes prefer unpaved roads over paved cycleways, which is not what I'd expect.

Example, road bike prefers an unpaved road:
https://maps.openrouteservice.org/directions?n1=52.213954&n2=6.843522&n3=17&a=52.214155,6.848248,52.216002,6.837723&b=1c&c=0&k1=en-US&k2=km

Mountain bike prefers cycleway (which I'd expect for road bike):
https://maps.openrouteservice.org/directions?n1=52.213954&n2=6.843522&n3=17&a=52.214155,6.848248,52.216002,6.837723&b=1b&c=0&k1=en-US&k2=km

marq24:
thanks for a reply...

So for me to clarify the (unwanted) unpaved section, that the road-bike route is that short way https://www.openstreetmap.org/way/136120460 ?!

I will look into that - in the meanwhile have you understood, why your provide code change will have other "unwanted" routes [for road bike cyclist] (the example I have provided (with the roundabouts is at a lot of places (and can IMHo currently be solved by giving cycle & foot ways (quite low speed)...

steventebrinke:
> in the meanwhile have you understood, why your provide code change will have other "unwanted" routes [for road bike cyclist] (the example I have provided (with the roundabouts is at a lot of places (and can IMHo currently be solved by giving cycle & foot ways (quite low speed)...

Well, if your example is the way road bikes cycle in Germany, then that's a country difference and I don't know how to solve that one, as there is no specific road bike tagging. In the Netherlands, cyclist would not be allowed to cycle on the main road when there is a cycleway next to it, so the main road is tagged with bicycle=no. This means that in cases similar to the one you point out, the planned route will rather go through the small village roads than along the main roads, even though that is not what most cyclists prefer.

steventebrinke:
> So for me to clarify the (unwanted) unpaved section, that the road-bike route is that short way https://www.openstreetmap.org/way/136120460 ?!

Indeed. That is where you'd not go with a road bike. With a road bike, you'd definitely prefer cycleways.

steventebrinke:
Another small example:
https://classic-maps.openrouteservice.org/directions?n1=52.236596&n2=6.840306&n3=18&a=52.235833,6.841875,52.236789,6.839182&b=1c&c=0&k1=en-US&k2=km

All bike types would take the cycleway here, but for road bike, the cycleway is discouraged, so it prefers steps.

marq24:
> Well, if your example is the way road bikes cycle in Germany, then that's a country difference and I don't know how to solve that one, as there is no specific road bike tagging. In the Netherlands, cyclist would not be allowed to cycle on the main road when there is a cycleway next to it, so the main road is tagged with bicycle=no. This means that in cases similar to the one you point out, the planned route will rather go through the small village roads than along the main roads, even though that is not what most cyclists prefer.

When you select the default "bike/cycle" profile you should receive the "legal conform" cycling route - the implemented road bike profile is not only targeting the avoidance of none compact surfaces (like gravel) - It's targeting 'common sense' road bike group ride routes...

I would never go though a roundabout with a road bike like it's suggested within the PR. Roads tagged with 'bicycle=no' should be also avoided by the current implemented "road bike' profile. So if a street is not permitted for bikes, then the routing should avoid it - if it's not explicit forbidden the routing will accept it. 

At the end of the day it's about point of views - when I summarize it: One point of view is: RoadBike will (just) avoid unpaved/gravel roads - but except of that you get the default cycle profile 1:1. The other (current implemented is) - stay as long as possible on regular roads (avoiding primary) - use cycleways only when there is no alternative...

And yes - since all is based on avg speeds - as short the route gets - as higher the chance is, that the routing might include invalid sections... [as in your last example - the long "cycleway" usage penalty is larger then the "use the steps" penalty]...  The good thing about ORS is, that it allows you to add additional via points to allow you to correct specific sections - there is no way based on the inconsequent tagging of the osm data to find a algorithm that will "always give you the optimal result" that you will choose as human (from knowing the area).

steventebrinke:
As you pointed out, I really like the ORS UI which easily allows adding via points, but as there is a limit on the number of via points, I cannot plan a day of cycling as a single route, as the normal bike profile takes smaller roads than I desire and the road bike profile is worse. If the goal is to avoid cycleways, IMHO it would be better if the profile was called "avoid cycleways" as "road bike" provides a different expectation. With a road bike in the Netherlands you'd usually take cycleways, especially on roundabouts like the ones you show, the cycleway would have right of way over cars, so you'd prefer the cycleway.

aoles:
Thanks @steventebrinke for your contribution, and @marq24 for raising a few important points in the comments! ❤️ 

The proper approach to implementing cycleway and footway avoidance would be through adjusted way priority values rather than decreased average speed values. Therefore, my suggestion is to mostly keep the changes to default speed value as proposed in this PR, but accompany them with corresponding changes to the priority values. This should address the routing issues mentioned in the examples form the discussion while keeping the overall road-bike routing experience.

aoles:
Moved to https://github.com/GIScience/openrouteservice/pull/2076 in order to be able to contribute to the PR.

PR #784, opened 2020-08-10
Allow configuration of app.config at docker run time (Fix #772)

EWouters:
This PR fixes #772.

I understand that you want the image to be able to run a demo out of the box, but also be usable as a base for custom app.configs.

In that case it would make sense to copy the app.config in the /ors-conf directory in the image to the $source_appconfig path if it exists.

This does not allow a user that does not supply an app.config to view the default app.config, but that use case is less important than the custom app.config use case in my opinion. This can easily be clarified in the documentation.

The logic is then:
```
# if Tomcat built before, copy the mounted app.config to the Tomcat webapp app.config, else copy it from the source
if [ -d "/usr/local/tomcat/webapps/ors" ]; then
	cp -f /ors-conf/app.config $tomcat_appconfig
else
	# if the user mounts an app.config file in /ors-conf, use the custom app.config
	if [ -f /ors-conf/app.config ]; then
		cp -f /ors-conf/app.config $source_appconfig
	fi
	echo "### Package openrouteservice and deploy to Tomcat ###"
	cp -f /ors-core/openrouteservice/target/*.war /usr/local/tomcat/webapps/ors.war
fi
```

I'll make a new PR for this fix.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -23,8 +23,9 @@ fi
 if [ -d "/usr/local/tomcat/webapps/ors" ]; then
 	cp -f /ors-conf/app.config $tomcat_appconfig
 else
-	if [ ! -f /ors-conf/app.config ]; then
-		cp -f $source_appconfig /ors-conf/app.config
+	# if the user mounts an app.config file in /ors-conf, use the custom app.config
+	if [ -f /ors-conf/app.config ]; then
+		cp -f /ors-conf/app.config $source_appconfig
 	fi
 	echo "### Package openrouteservice and deploy to Tomcat ###"
 	mvn -q -f /ors-core/openrouteservice/pom.xml package -DskipTests && \


rabidllama:
Could there not be an else statement that copies the used `app.config` to the `ors-conf` folder if there isn't one present? Then it would be a case that if there is one present, it overwrites the generated one, but if there isn't one then the "default" config is saved for the user to edit in future runs

EWouters:
That could work but I don't think that would be clear to users.

In that case they can start a container from the image without customizing an app.config file, but with mounting the conf folder. They get a default instance then and the app.config file is copied to the conf folder. The app.config file will be owned by root.
They can then change the ownership of the file and make edits. When they try to start the container again the name will be taken (if they specified a name) so they will have to remove the container of that name and recreate it to use their custom app.config.

I think users that want a custom setup will be able to download the app.config file from the documentation, edit it and place it in the conf folder and mount it, much like the .osm file.

This removes some confusion about starting 2 different containers and file ownership.

rabidllama:
Yes I understand that file ownership can be a bit confusing (thanks Docker...). I think the problem is that the sample config file used when none is provided is edited during the build process to take into account the internal container file structure, and the reason that was done was to remove the redundancy in the amount of configs that have to be updated whenever a change is made. 

Do you think that doing as you suggest and not writing out the default to `app.config` but instead to one called `app.config.base` (or even `app.config.sample`) or similar would be an option? Sometimes it is the case that debugging is a lot easier if the user can see the config file that was used and don't have to work out what the Dockerfile did or go into the container to extract the config file used.

rabidllama:
Did you have any further thoughts about this @EWouters ?

rabidllama:
As we haven't had any other feedback about the questions, I will close this PR for now without a merge

PR #799, opened 2020-08-28
fix handling of invalid extra info requests

### LINKED ISSUE #795: Incomplete extra info returned when using items invalid for profile
#### Here's what I did
Request a route with multiple extra info items chosen, some of which are not valid for the selected profile
https://api.openrouteservice.org/v2/directions/foot-walking/json?api_key=xxx
```json
{
	"preference":"recommended",
	"coordinates":[[9.736805,47.100863],[9.783754,47.078656]],
	"extra_info":["steepness","suitability","surface","waycategory","waytype","tollways","traildifficulty","osmid","roadaccessrestrictions","countryinfo","green","noise"]
}
```
---
#### Here's what I got
Not all valid and chosen extra info items are returned (in the above case only steepness, suitability, surface, waycategory and waytype were returned. 

---
At least the traildifficulty should have been returned. When "invalid" items are removed, the response comes back as expected (all selected extra info items are returned in the respnce:
```json
{
	"preference":"recommended",
	"coordinates":[[9.736805,47.100863],[9.783754,47.078656]],
	"extra_info":["steepness","suitability","surface","waycategory","waytype","traildifficulty","green","noise"]
}
```

---
The response should always include the extra info items chosen, unless they are not valid for that profile

takb:
### Pull Request Checklist

Fixes #795 .

- Key functionality added:
Handles invalid extra info requests (missing graph extensions meaning extra info not available for the requested profile) by adding a warning and continuing processing of other extra info requests rather than stopping extra info processing and only logging a useless error message on the server.

### DIFF
--- a/openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfileManager.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfileManager.java
@@ -13,8 +13,8 @@
  */
 package org.heigit.ors.routing;
 
+import com.google.common.base.Strings;
 import com.graphhopper.GHResponse;
-import com.graphhopper.routing.util.EncodingManager;
 import com.graphhopper.util.DistanceCalc;
 import com.graphhopper.util.Helper;
 import com.graphhopper.util.PointList;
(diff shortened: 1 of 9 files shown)


openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfileManager.java
```diff
@@ -13,8 +13,8 @@
  */
 package org.heigit.ors.routing;
 
+import com.google.common.base.Strings;
```
rabidllama: Wouldn't it be better to use internal Java checks for null and empty string rather than including a google library?

rabidllama (changes requested):
Mostly all good, just a query regarding the use of a google library

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfileManager.java
takb: replaced all usages of `com.google.common.base.Strings` with a new method in `org.heigit.ors.util.StringUtility`.

rabidllama (approved):

PR #813, opened 2020-09-25
Bugfix#683 storagebuilders init

### LINKED ISSUE #683: Setting elevation : false while still building HillIndex storage breaks creation of other following extended storages.
The function
```public void processEdge(ReaderWay way, EdgeIteratorState edge, Coordinate[] coords) {
		if(arrStorageBuilders != null) {
			int nStorages = arrStorageBuilders.length;
			for(int i=0; i<nStorages; i++) {
				arrStorageBuilders[i].processEdge(way, edge, coords);
			}
		}
	}
```
in GraphProcessContext contains a HillIndexStorage at arrStorageBuilder[0] if elevation is set to false in the app.config, while the HillIndexStorage is indicated in the extended storages. 
The init of this storage is potentially not (correctly) called and results in a HillIndexGraphStorageBuilder that has a field hillIndexCalc == null.
This leads to an early termination of the above function, which is caught with only a warning in the execution of the above code block. As the HillIndexStorageBuilder is at array position 0, none of the following StorageBuilders are called, while no proper exception is being thrown.
Removing the HillIndexStorage from the extended storages in the app.config solves the problem, but the failing combination should be caught in the code.
The init of the storage should in theory already return a null Builder for this storage given no elevation, but it does not seem to work. Maybe something broken with graphhopper.hasElevation()?
Potentially due to app.config parameter  ` elevation_provider: multi `being set

takb:
### Pull Request Checklist

Fixes #683  .

- Reason for change:
Correctly handle the HillIndexStorage not being able to initialize (if elevation data missing). Also added checks and feedback during ExtraInfoProcessor instantiation. 
Also fixed: Setting the elevation parameter to false for a profile was ignored if previous profiles had it enabled,

### DIFF
--- a/openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
@@ -176,7 +176,9 @@ public class RoutingProfile {
         // so the caching for elevation data will/can be reused across different vehicles. [the loadCntx is a single
         // Object that will shared across the (potential) multiple running instances]
         if(loadCntx.getElevationProvider() != null) {
-            gh.setElevationProvider(loadCntx.getElevationProvider());
+            if (args.has("graph.elevation.provider")) {
+                gh.setElevationProvider(loadCntx.getElevationProvider());
+            }
         }else {
             loadCntx.setElevationProvider(gh.getElevationProvider());
         }
--- a/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
@@ -123,7 +123,6 @@ public class ORSGraphHopper extends GraphHopper {
 		fastIsochroneFactory.init(args);
 		minNetworkSize = args.getInt("prepare.min_network_size", minNetworkSize);
 		minOneWayNetworkSize = args.getInt("prepare.min_one_way_network_size", minOneWayNetworkSize);
-
 		return ret;
 	}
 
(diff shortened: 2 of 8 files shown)


openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
```diff
@@ -176,7 +176,9 @@ public static ORSGraphHopper initGraphHopper(String osmFile, RouteProfileConfigu
         // so the caching for elevation data will/can be reused across different vehicles. [the loadCntx is a single
         // Object that will shared across the (potential) multiple running instances]
         if(loadCntx.getElevationProvider() != null) {
-            gh.setElevationProvider(loadCntx.getElevationProvider());
+            if (args.has("graph.elevation.provider")) {
```
HendrikLeuschner, line 179: What happens if `(loadCntx.getElevationProvider() != null) && !args.has("graph.elevation.provider")`

takb: the loadcontext is used to store elevationprovider if it has been instantiated before, as far as i understand to make enable caching of data, and the problem was that every later profile was assigned the provider regardless if it had elevation parameter set to true. in the case in question the next line is supposed to not run.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
```diff
@@ -141,6 +140,7 @@ protected void cleanUp() {
 			if (LOGGER.isInfoEnabled())
 				LOGGER.info(String.format("will create PrepareRoutingSubnetworks with:%n\tNodeCountBefore: '%d'%n\tgetAllEdges().getMaxId(): '%d'%n\tList<FlagEncoder>: '%s'%n\tminNetworkSize: '%d'%n\tminOneWayNetworkSize: '%d'", prevNodeCount, ex, list, minNetworkSize, minOneWayNetworkSize)
 			);
+			ghs.getProperties().put("elevation", hasElevation());
```
HendrikLeuschner, line 143: What is this used for? In the HillIndexStorage, graphhopper.hasElevation() is called directly.

HendrikLeuschner (approved):
Looks fine to me

PR #815, opened 2020-10-08
Update to GraphHopper ors_v0.13.8

aoles:
### Pull Request Checklist

- Key functionality added:
Introduce time-dependent Dijkstra and A* into ORS.
- Reason for change:
Base for for time-dependent isochrone or route calculation.


HendrikLeuschner:
Why are the param tests getting a connection refused error?

takb:
my bad, messed up the gh release (tagged the wrong branch...), I'm working on fixing it.

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
```diff
@@ -989,7 +1003,8 @@ private void setWeighting(HintsMap map, int requestWeighting, int profileType){
         //For a requested recommended weighting, use recommended for bike, walking and hgv. Use fastest for car.
         if (requestWeighting == WeightingMethod.RECOMMENDED || requestWeighting == WeightingMethod.FASTEST) {
             if (profileType == RoutingProfileType.DRIVING_CAR) {
-                weighting = weightingMethod = VAL_FASTEST;
+                weighting = VAL_FASTEST;
```
HendrikLeuschner, line 1009: Do we only support car with td? Also can we make "td_fastest" a String variable like the others?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
```diff
@@ -361,6 +361,29 @@ else if (ALT_ROUTE.equalsIgnoreCase(algoStr))
 				if (weighting instanceof TurnWeighting)
 	                ((TurnWeighting)weighting).setInORS(true);
 
+				weighting = createTimeDependentAccessWeighting(weighting, tMode, algoStr);
```
HendrikLeuschner: Is it correct to enforce the time dependent weighting to be an access weighting here? should it not be something more general?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphStorageFactory.java
```diff
@@ -86,9 +86,6 @@ else if (geTurnCosts != null && !graphExtensions.isEmpty()) {
 				((ORSGraphHopper) gh).initCoreLMAlgoFactoryDecorator();
 		}
 
-		if (gh.getLMFactoryDecorator().isEnabled())
```
HendrikLeuschner, line 91: Why was this removed?

aoles: Fair point, this is quite obscure. This change was complemented by https://github.com/GIScience/graphhopper/commit/b44430e3ded517b4f161bd5ff16e7c495c04a1a1 and the rearrangement was necessary in order to get things to work. Unfortunately, I don't remember the exact details anymore, but if you follow the program flow then you will realize that before this change we had to call `initLMAlgoFactoryDecorator();` only because the original call in GH never happened because of this very modification in the fork:

```//ORS-GH MOD START
        if (graphStorageFactory != null) {
            ghStorage = graphStorageFactory.createStorage(dir, this);
        }

        if (ghStorage == null) {
        //ORS-GH MOD END```

HendrikLeuschner: Okay, seems reasonable to me. The order of initialization is also kept intact. As long as the code in ORSGraphStorageFactory that follow the initialization of core and corelm does not depend on LM, I do not see a reason why we should not do it on the "outer layer". But why is the CH part still in this factory and not in the graphhopper code?

aoles: The CH part has never been a subject to time-dependent routing and as such has not been modified in these PRs. 

The change to LM creation was necessary because of the fact that currently `TimeDependentFastestWeighting` requires `GraphHopperStorage` to be initialized beforehand in order to create the `SpeedCalculator`. The original code arrangement failed with a `NullPointerException`.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
aoles: This call is supposed to wrap any super `weighting` (say `shortest`, `fastest`, or `td_fastest` into a `TimeDependentAccessWeighting`  similarly as `createTurnWeighting`, so I don't really see an issue here.

HendrikLeuschner: I understand that, I was wondering whether this would be correct under any circumstance, or whether a generic td weighting would be better. This generic weighting can then be made an accessWeighting if the conditions (I do not know whats necessary for that) are met

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
aoles: Thanks for these suggestions! The logic was adapted from the original tardur implementation which indeed has the td routing enabled just for car and bike encoders. I can look into improving this part.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
aoles: So as long as time-dependent routing is enabled and a conditional access storage is available to a given encoder it makes sense to have `TimeDependentAccessWeighting` enabled for the sake of correctness of routes. Do you have any particular time-dependent weighting in mind which would fit into this scheme? When it comes to the fastest-based weightings: it doesn't seem feasible to me to have them implemented as wrapper weightings because of the conversion from speed to the time domain which happens in them internally.

HendrikLeuschner: I do not have one in mind. But we might come to that point later. Is the AccessStorage always available? Are there circumstances where it is not created?

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
aoles: I've made some adaptations. Actually `td_fastest` has been implemented and tested just with the car profile, so I have kept that logic. This part will be updated anyway once we introduce proper td routing.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
aoles: There is currently no mechanism implemented which would prevent it from being created.

aoles: OK, seems that it would be reasonable to add to `createTimeDependentAccessWeighting` method a check for the presence of `conditional_access` encoded value for a given flag encoder. I will fix this.

aoles: Done!

HendrikLeuschner (approved):
Looks fine to me now.

PR #837, opened 2020-12-04
Fi optimize active cells

HendrikLeuschner:
### Pull Request Checklist

- [] 5. I have written JUnit tests for any new methods/classes and ensured that they pass.

Fixes # .

- Key functionality added:Optimization in Fast isochrone runtim calculation of active cells
- Reason for change: its faster now

-Fixed geometry calculation for small intervals in small isochrones.

-

-


HendrikLeuschner:
Need to fix an additional bug first

HendrikLeuschner:
From my side ready to merge. Did 10000 tests on Germany with no failure.

openrouteservice/src/main/java/org/heigit/ors/fastisochrones/CoreRangeDijkstra.java
```diff
@@ -68,8 +68,6 @@ protected void runAlgo() {
                 int traversalId = traversalMode.createTraversalId(iter, false);
                 // Modification by Maxim Rylov: use originalEdge as the previousEdgeId
                 double tmpWeight = weighting.calcWeight(iter, reverseDirection, currEdge.originalEdge) + currEdge.weight;
-                if(tmpWeight > isochroneLimit)
```
aoles, line 71: Why has been this check removed?

openrouteservice/src/main/java/org/heigit/ors/fastisochrones/Contour.java
```diff
@@ -44,7 +44,7 @@
     private static final int SUPER_CELL_HIERARCHY_LEVEL = 3;
     //This means that one supersupercell can contain at most 2^(3 + 2) = 32 base cells.
     private static final int SUPER_SUPER_CELL_HIERARCHY_LEVEL = 2; // level above super cell level
-    private static final double CONCAVE_HULL_THRESHOLD = 0.010;
+    private static final double CONCAVE_HULL_THRESHOLD = 0.006;
```
aoles, line 47: What does this magic number control exactly and why has it been changed from `0.010` to `0.006` and not for example `0.005`?

aoles (changes requested):
Overall looks good to me - just a couple of questions, see below.

openrouteservice/src/main/java/org/heigit/ors/fastisochrones/CoreRangeDijkstra.java
HendrikLeuschner: Keeping this change means that any edge whose length + already visited weight is greater than the limit will not be put into the map. This means that the limit will not be fully reached, even if a part of the edge would still be inside the limit.
There is a mechanism however in the later processing of the map, that will subdivide edges that are half in, half out of the isolimit by looking at their actual geometry. Removing this check here means that those edges will be included in the map and can be properly processed. Makes it more correct. See the `finished` condition.

openrouteservice/src/main/java/org/heigit/ors/fastisochrones/Contour.java
HendrikLeuschner: This changes the precision with which the concave hull of the contour is outlined. This happens in preprocessing. The precision was adapted to be higher (more detail), choosing the number which is equal to the highest possible precision that can be used during actual query processing.
By using the `smoothing` factor in the query, the user can define a more detailed isochrone. If the contours have been preprocessed with less precision, it leads to inconsistent behavior (fully reachable cells are less detailed than active cells) and in the worst case to the "break-in" of the concave hull algorithm into the contour, as now the distance between two contour nodes is too large for the high precision query time algorithm to consider it a connection, thus sort of hollowing out the whole cell from the inside. This is hard to explain here.

openrouteservice/src/main/java/org/heigit/ors/fastisochrones/CoreRangeDijkstra.java
aoles: Makes sense, thanks for clarifying!

openrouteservice/src/main/java/org/heigit/ors/fastisochrones/Contour.java
aoles: No worries, I was just curious how you came up with this very value.

aoles (approved):
We're good to go!

HendrikLeuschner:
Thanks!

takb:
Sorry guys, the CI test isn't working the way it's supposed to yet, it's reporting pass while there are api-tests failing. The following errors occur on the latest version: 

[ERROR]   ResultTest.testAreaUnitsOverridesUnits:292 1 expectation failed.
JSON path features[0].properties.area doesn't match.
Expected: is (a value greater than <12.1F> and a value less than <12.2F>)
  Actual: 12.057727

[ERROR]   ResultTest.testIntersections:349 1 expectation failed.
JSON path features[2].geometry.coordinates[0].size() doesn't match.
Expected: is <38>
  Actual: 41

[ERROR]   ResultTest.testPolygon:119 1 expectation failed.
JSON path features[0].geometry.coordinates[0].size() doesn't match.
Expected: is <40>
  Actual: 43

[ERROR]   ResultTest.testReachfactorAndArea:215 1 expectation failed.
JSON path features[0].properties.area doesn't match.
Expected: is (a value greater than <1.21E7F> and a value less than <1.22E7F>)
  Actual: 1.2057726E7

[ERROR]   ResultTest.testReachfactorAndAreaAreaUnitsKM:265 1 expectation failed.
JSON path features[0].properties.area doesn't match.
Expected: is (a value greater than <12.1F> and a value less than <12.2F>)
  Actual: 12.057727

[ERROR]   ResultTest.testReachfactorAndAreaAreaUnitsM:240 1 expectation failed.
JSON path features[0].properties.area doesn't match.
Expected: is (a value greater than <1.21E7F> and a value less than <1.22E7F>)
  Actual: 1.2057726E7

[ERROR]   ResultTest.testReachfactorAndAreaAreaUnitsMI:316 1 expectation failed.
JSON path features[0].properties.area doesn't match.
Expected: is (a value greater than <4.68F> and a value less than <4.69F>)
  Actual: 4.6555123

I have verified that before merging current master (with @aoles ' conditional access fix) they were not there. I assume they are only cases where the result values have to be adjusted, but you're going to have to have a look @HendrikLeuschner ...

HendrikLeuschner:
Sure, thanks. I thought I had already adapted the values for all the changed tests, but apparently not.

PR #849, opened 2021-01-05
fix runfile error handling

### LINKED ISSUE #832: Redefine automated tests using github action
- [x] set up github action
- [x] get rid of travis
- [x] fix chat-notification

takb:
### Pull Request Checklist

Fixes #832 .

Fixed error handling for the case writing ors.run file fails, added note to changelog.

### DIFF
--- a/.github/workflows/project_automation.yml
+++ b/.github/workflows/project_automation.yml
@@ -12,7 +12,7 @@ jobs:
   Handle_event:
     runs-on: ubuntu-latest
     steps:
-      - name:
+      - name: run project board action
         uses: takb/project-board@master
         env:
           PROJECT_AUTOMATION: ${{ secrets.PROJECT_AUTOMATION }}
--- a/.github/workflows/project_automation.yml
+++ b/.github/workflows/project_automation.yml
@@ -12,7 +12,11 @@ jobs:
   Handle_event:
     runs-on: ubuntu-latest
     steps:
-      - uses: takb/project-board@master
+      - name:
+        uses: takb/project-board@master
+        env:
+          PROJECT_AUTOMATION: ${{ secrets.PROJECT_AUTOMATION }}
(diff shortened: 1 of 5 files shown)


.github/workflows/project_automation.yml
```diff
@@ -12,7 +12,11 @@ jobs:
   Handle_event:
     runs-on: ubuntu-latest
     steps:
-      - uses: takb/project-board@master
+      - name:
```
rabidllama: Should there be a name here?

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfileManager.java
```diff
@@ -625,12 +625,12 @@ public MatrixResult computeMatrix(MatrixRequest req) throws Exception {
...
 
     public void createRunFile() {
-        File file=new File("ors.run");
+        File file = new File("ors.run");
         try (FileWriter fw = new FileWriter(file)) {
             fw.write("ORS init complete: "+ Instant.now().toString() + "\n");
             fw.flush();
-        } catch(IOException ex) {
-            ex.printStackTrace();
+        } catch(Exception ex) {
```
rabidllama, line 632: Isn't it better to catch an explicit exception rather than the generic one (Sonar lint always brought that up for me anyway)?

takb: normally yes, but in this case I don't see the reasoning since it's just a "ignore any errors" and it would require several different catch blocks for the same thing which is simply annoying...

.github/workflows/project_automation.yml
takb: you're right, though it really doesn't matter...

rabidllama (approved):

PR #863, opened 2021-02-23
fix missing instructions if geometry is set to "false"

### LINKED ISSUE #851: Instructions missing if Geometry is set to False
I’am calling directions with POSt and using this like body

```
{
   "preference":"shortest",
   "units":"km",
   "language":"fr",
   "geometry":"false",
   "instructions":"true",
   "coordinates":[
      [
         6.791700,
         51.999200
      ],
      [
         6.791700,
         53.109200
      ]
   ],
   "options":{
      "avoid_countries":[
         
      ],
      "avoid_borders":"none",
      "avoid_features":[
         
      ],
      "profile_params":{
         "restrictions":{
            "width":0.0,
            "height":0.0,
            "weight":32.0,
            "length":0.0,
            "axleload":3,
            "hazmat":false
         }
      }
   }
}
```

I went to have the response **without geometry**, but even with geometry = false I’am having it on my response. The only way to remove it, is to put **instructions and geometry both to false**, but this will also **remove the segments and the instructions from the response**.

Are there anyway **to have the instructions without the geometry** when i’am routing only two points?

PS: When having more than two points and geometry false and instructions true it worked, i can have the instructions without the geometry.

koebi:
### Pull Request Checklist

Fixes #851.

However, as mentioned there, tests should be added to make sure such an issue doesn't appear again.


takb (approved):
I guess an API test can be added some time later...

PR #866, opened 2021-03-02
Enable turn restrictions in CALT routing algorithm

### LINKED ISSUE #807: not adhering to turn restrictions without alternative routes
#### Here's what I did

Strange behaviour when not having alternative routes:

```
  curl -X POST \
  'https://api.openrouteservice.org/v2/directions/driving-car/geojson' \
  -H 'Content-Type: application/json; charset=utf-8' \
  -H 'Accept: application/json, application/geo+json, application/gpx+xml, img/png; charset=utf-8' \
  -H 'Authorization: xxx' \
  -d '{"coordinates":[[10.952897,48.929647],[10.971179,48.931508]]}'
```
---
#### Here's what I got

grafik

---

When setting alternative routes it strangely shows the correct route:
```
  curl -X POST \
  'https://api.openrouteservice.org/v2/directions/driving-car/geojson' \
  -H 'Content-Type: application/json; charset=utf-8' \
  -H 'Accept: application/json, application/geo+json, application/gpx+xml, img/png; charset=utf-8' \
  -H 'Authorization: xxx' \
  -d '{"coordinates":[[10.952897,48.929647],[10.971179,48.931508]],"alternative_routes":{"target_count":2}}'
```

grafik

star-pirate:
### Pull Request Checklist

Fixes #807 .

- Key functionality added: support for turn restrictions with core routing algorithms
- Reason for change: enforce turn restrictions in CoreALT 

-

-


openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/AbstractCoreRoutingAlgorithm.java
```diff
@@ -37,29 +40,42 @@
...
+
+    public AbstractCoreRoutingAlgorithm(Graph graph, Weighting weighting) {
+        super(graph, new PreparationWeighting(weighting), TraversalMode.NODE_BASED);
+
+        if (weighting instanceof TurnWeighting) {
+            turnWeighting = (TurnWeighting) weighting;
+            hasTurnWeighting = true;
+        }
+        else {
+            turnWeighting = null;
```
HendrikLeuschner: Is this necessary if there is `hasTurnWeighting` and `turnWeighting` is initialized to `null` already?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/AbstractCoreRoutingAlgorithm.java
```diff
@@ -160,11 +172,42 @@ public RoutingAlgorithm setEdgeFilter(CoreDijkstraFilter additionalEdgeFilter) {
...
         return this;
     }
-    protected boolean accept(EdgeIterator iter, int prevOrNextEdgeId) {
-        return additionalCoreEdgeFilter == null || additionalCoreEdgeFilter.accept(iter);
+
+    //TODO: refactor CoreEdgeFilter to plain EdgeFilter to avoid overriding this method
+    @Override
+    protected boolean accept(EdgeIteratorState iter, int prevOrNextEdgeId) {
+        if (iter.getEdge() == prevOrNextEdgeId) {
+            return false;
```
HendrikLeuschner, line 175: Is this for u turns? Are u turns (the restriction of u turns) fully supported now?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/AbstractCoreRoutingAlgorithm.java
```diff
@@ -160,11 +172,42 @@ public RoutingAlgorithm setEdgeFilter(CoreDijkstraFilter additionalEdgeFilter) {
...
+    }
+
+    boolean isTurnRestrictedNode(int node) {
+        return chGraph.getLevel(node) == turnRestrictedNodeLevel;
+    }
+
+    boolean considerTurnRestrictions(int node) {
+        if (!hasTurnWeighting)
+            return false;
+        if (approximate)
```
HendrikLeuschner, line 203: Is the approximation able to be set by config? Whats the implications of it exactly now? also performance wise

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
```diff
@@ -229,7 +245,7 @@ void contractNodes() {
                 sortedNodes.clear();
                 int len = prepareGraph.getNodes();
                 for (int node = 0; node < len; node++) {
-                    if (prepareGraph.getLevel(node) != maxLevel)
+                    if (prepareGraph.getLevel(node) < maxLevel)
```
HendrikLeuschner: `!isCoreNode(node)`?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
```diff
@@ -268,31 +284,17 @@ void contractNodes() {
...
-
             if (sortedNodes.getSize() < nodesToAvoidContract) {
                 // skipped nodes are already set to maxLevel
                 prepareGraph.setCoreNodes(sortedNodes.getSize() + 1);
                 //Disconnect all shortcuts that lead out of the core
                 while (!sortedNodes.isEmpty()) {
                     CHEdgeIterator iter = vehicleAllExplorer.setBaseNode(polledNode);
                     while (iter.next()) {
-                        if (prepareGraph.getLevel(iter.getAdjNode()) == maxLevel) continue;
+                        if (prepareGraph.getLevel(iter.getAdjNode()) >= maxLevel) continue;
```
HendrikLeuschner: `isCoreNode(iter.getAdjNode())`?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
```diff
@@ -365,7 +367,21 @@ void contractNodes() {
                 + ", " + Helper.getMemInfo());
     }
 
-
+    private void setTurnRestrictedLevel(int polledNode) {
+        CHEdgeIterator edge1 = vehicleAllExplorer.setBaseNode(polledNode);//FIXME: iterate only over original edges no need to consider shortcuts?
```
HendrikLeuschner: Whats with this FIXME?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
aoles: Thanks for the suggestion! At the beginning of the algorithm all nodes are set to `maxLevel` and assigned a lower level only when contracted. Even though a lower level effectively means that the node won't be included in the core, in this particular context what matters more is whether the node has been already contracted or not. I'm afraid that using a function call `!isCoreNode` to check for this instead of a direct comparison might be a bit confusing and make it more difficult to read and to understand the code.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
aoles: As far as I can tell this is literally the one and only place where the check for a node being in the core is performed, so I'm not entirely sure whether it makes sense to abstract it to a function. As mentioned in my previous comment, considering the algorithm context the fact of a node being at level `< maxLevel` should rather be interpreted as being contracted rather than not being a core node. Therefore I'm a bit hesitant to recycle such a function for a conceptually different purpose.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/AbstractCoreRoutingAlgorithm.java
aoles: That's correct, this check is necessary to prevent u-turns.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/AbstractCoreRoutingAlgorithm.java
aoles: Both `turnWeighting` and `hasTurnWeighting` have been originally declared as `final`. As such they needed to be initialized with `null` and `false`, respectively. Because it is not critical for them to be `final` I guess we can omit this for the sake of simpler code.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/AbstractCoreRoutingAlgorithm.java
aoles: It is currently not possible to use the approximated algorithm other than by setting the corresponding flag in the source code. This is on purpose so and there are various reasons for it. First of all, once the approximation is enabled the resulting route for a given weighting is no longer guaranteed to be the optimal one. This is different from the rest of the algorithms we use which are exact in those terms. The reason for this is mostly because alternative paths involving edges which have been reached previously on a shorter way are not explored anymore. This can happen, for example,  when the forward or backward search enters a street from both ends. Once a node on such street is relaxed it would not be considered anymore, even though because of turn restrictions the longer path going in the opposite direction might actually be eventually involved in the shortest path in question. This is solved by the GH's original edge-based algorithm by distinguishing between the direction of the edges. In order to reproduce this in node-based routing one needs to keep a list of weights corresponding to the incoming edges for each and every node. The approximated algorithm does this only for nodes involved in turn relations. Although some initial testing suggests that the resulting differences tend to be small and occur in < 1% of the routes, it merits a more thorough assessment in order to better understand its practical implications. But more importantly, while performance-wise it makes quite a difference for CoreDijkstra, I was not able to observe any significant impact on CoreALT, at least on the Berlin dataset. There was virtually no measurable difference in query times between the exact and the approximate solution (I shall upload the results on Confluence). Therefore, I've decided to keep the implementation dormant with the option to easily switch it on in the source code in case we are interested in pursuing this idea in the future.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
aoles: Right, it's more of a todo-type of thing. It's correct as it is now, but because turn relations exist only between original edges it is not necessary to include shortcuts in the iterator. I will look into improving it.

aoles: This has been addressed in https://github.com/GIScience/openrouteservice/pull/866/commits/08aae57583025779e77f8037332585d60fc94958.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/AbstractCoreRoutingAlgorithm.java
HendrikLeuschner: Cool. So no more discrepancies between ALT and CALT concerning u turns now? Or are there some fringe cases still?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/AbstractCoreRoutingAlgorithm.java
HendrikLeuschner: Thanks for the explanation, sounds reasonable to me given that the performance with CALT is no different.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
HendrikLeuschner: Understandable. I think from a maintenance perspective, it is still useful to extract a method that has a meaningful name so that someone who reads this can understand what is going on here. Does not have to be `isCoreNode`, just something that explains what this inequality means. But its a minor complaint.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/PrepareCore.java
HendrikLeuschner: Thx

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/AbstractCoreRoutingAlgorithm.java
aoles: That's the case as far as I can tell. The remaining discrepancies [listed on confluence](https://confluence.gistools.geog.uni-heidelberg.de/x/hYHhAw) in seem to be of different nature.

HendrikLeuschner (approved):

PR #870, opened 2021-03-08
fix SONAR_TOKEN conditional

takb:


aoles (commented):
Thanks! Wouldn't it be cleaner to run just the API tests in the first step regardless of the sonar token status, and try the sonar checks only in the next step? The described approach seems a bit less redundant to me and therefore cleaner..

takb:
you're right, i'f sort of confused today...

aoles (approved):
Thanks!

PR #871, opened 2021-03-15
fix distance limit calculation with skipped segments

### LINKED ISSUE #814: Skipped segments should not count on the distance limit
API returns limit exceeded error even when the whole segment is marked to be skipped. 

If a segment is to be skipped, then it should not count on the distance limit.

Request 
URL: `https://api.openrouteservice.org/v2/directions/driving-car`
body: `{"coordinates":[[8.681495,49.41461],[-38.43326568603516, -12.967758119178905]],"skip_segments":[1]}`

Response:
`{"error":{"code":2004,"message":"Request parameters exceed the server configuration limits. The approximated route distance must not be greater than 6000000.0 meters."}}`

takb:
### Pull Request Checklist

Fixes #814.

- Reason for change:
Approximated distance calculation to limit requests is now ignoring the skipped segments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ RELEASING:
 ## [Unreleased]
 ### Added
 - API endpoint "centrality" to calculate [betweenness centrality](https://en.wikipedia.org/wiki/Betweenness_centrality) values for nodes inside a given bounding box. Centrality is calculated using Brandes' algorithm. 
+### Fixed
+- Fixed calculation of route distance limits with skipped segments ([#814](https://github.com/GIScience/openrouteservice/issues/814))
+
 ## [6.3.6] - 2021-02-02
 ### Fixed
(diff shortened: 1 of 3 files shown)


sfendrich (approved):

rabidllama (commented):
Everything looks good, but one thought:

Should the documentation for the API be updated to state that the distance returned does not include "skipped segments"? Something like "Total calculated route distance in specified units (ignoring skipped segments)." 

https://github.com/GIScience/openrouteservice/blob/cb6966cce4d9ebaaaf6d93b4dfc98de094e2561b/openrouteservice/src/main/java/org/heigit/ors/api/responses/routing/json/JSONSummary.java#L28

Not sure if there is also something for that in the segment information as well.

takb:
The distances of skipped segments are included in the result, the approximated total distance calculated here is only used for the preliminary max distance check.

PR #872, opened 2021-03-16
Issue 867 fix

### LINKED ISSUE #867: Isochrone snaps to wrong point
Isochrone requests only snap to actual edge tower nodes, they do not produce virtual nodes on the graph. That results in unexpected behavior in case the edge is very long

#### Here's what I did
e.g. https://bit.ly/388UhSB : In this example, there is an underlying tunnel that the request snaps to. This is a long edge and the algorithm snaps to the one at the western end, leading to the request point being outside of the actual isochrone. Why does it snap to this edge? Because the "find closest edge" algorithm takes into account the edge geometry, where the tunnel is actually the closest to the start point. But then it uses the original edge tower nodes as isochrone start points.

Edit: This same problem arises in routing, as I suspected: [Routing in tunnel](https://maps.openrouteservice.org/#/directions/Bismarckplatz,Essen,Germany/Bert-Brecht-Stra%C3%9Fe,Essen,Germany/data/55,130,32,198,15,97,4,224,38,9,96,59,2,24,5,192,166,6,113,0,184,64,118,0,232,0,102,32,54,1,152,2,96,17,128,86,0,88,4,224,126,186,0,226,96,26,58,108,33,230,53,201,11,167,76,131,50,236,3,113,21,45,65,190,124,21,243,183,96,206,149,42,60,248,10,102,190,139,98,237,240,209,5,196,4,0,14,169,224,68,77,135,40,75,208,32,3,55,128,6,221,46,16,177,163,192,3,114,64,7,48,5,163,6,70,131,49,6,119,69,119,70,135,68,67,6,243,196,76,128,5,176,202,77,135,69,134,136,2,242,128,205,193,167,96,5,247,42,0,0) This should instead find the upper road that is not a tunnel. But at least here the snapping works correctly, using a virtual node!

---
#### Here's what I got

---
Generally, the isochrones algorithm should generate virtual edges and virtual tower nodes from snapping in order to prevent this, just like it is done for routing. In this specific case, though, this would not make much sense. When you take the start point and pull it 20m north so that it is in the building with nmbr 16, the snapping should find the closest _accessible_ edge, and not the tunnel.
So even resolving the original issue, the result would still be unexpected for the user. This seems to be a fringe case though, as most edges are accessible as long as they are not physically separated. This might be a secondary issue.

---
First, we need the isochrones algorithm to create virtual nodes and take into account the snapping distance to those when calculating the isochrone extend. Second, the search should then start from those virtual nodes.

The alternative solution might be to modify the _findClosest_ algorithm so that it does not use edge geometry and instead only searches for the closest real tower node. This should stop it from finding the tunnel and instead find a node in the vicinity that seems more reasonable, like the parking space in the above example. The problem with this is that it will not work when you start in the middle of a very long edge where there are no alternatives around, then the case from the example persists.

HendrikLeuschner:
### Pull Request Checklist

Fixes #867  .

- Key functionality added: Use the querygraph that creates virtual nodes to snap to the closest edge instead of using the closest real node.
- Reason for change: Otherwise, the queried location may lie outside the isochrone.

- See issue

-


aoles (approved):
LGTM, thanks!

PR #874, opened 2021-03-18
include distance, duration and steps in all route segments in API res…

### LINKED ISSUE #695: Missing segment distance for POST request
#### Here's what I did

Running Openrouteservice using Docker on current `master`, I'm sending a POST request for directions where the two first coordinates are identical: `{"coordinates":[[8.694724,49.409358],[8.694724,49.409358],[8.669243,49.444468]]}`.

---
#### Here's what I got

Focusing on the first segment of the route (filtering with `jq`):

```bash
$ curl -X POST --header "Content-Type:application/json" -d '{"coordinates":[[8.694724,49.409358],[8.694724,49.409358],[8.669243,49.444468]]}' http://localhost:8080/ors/v2/directions/foot-walking | jq '.routes[0].segments[0]'
```
```json
{
  "steps": [
    {
      "distance": 0,
      "duration": 0,
      "type": 11,
      "instruction": "Head east",
      "name": "-",
      "way_points": [
        0,
        1
      ]
    }
  ]
}

```

---

I'm expecting `distance` and `duration` keys (with 0 values) for that segment. Note that it is the behaviour when running the same request with a GET:

```bash
curl "http://localhost:8080/ors/directions?profile=foot-walking&coordinates=8.694724,49.409358|8.694724,49.409358|8.669243,49.444468" | jq '.routes[0].segments[0]'
```
```json
{
  "distance": 0,
  "duration": 0,
  "steps": [
    {
      "distance": 0,
      "duration": 0,
      "type": 11,
      "instruction": "Head east",
      "name": "",
      "way_points": [
        0,
        1
      ]
    }
  ]
}
```

Except for the `name` values and stuff related to different format between POST and GET responses, this is the only change in the response.

---

Not sure if the current API always guarantee having a `distance` and `duration` key at `segment` level (even if values are zero), but having a consistent response across request types would be safer. Also this breaks [downstream code](https://github.com/VROOM-Project/vroom/blob/fed2ca8b52b687cddabdf9ca77582af54953ec6f/src/routing/ors_wrapper.cpp#L83) that goes through all segments, relying on `distance` keys.

koebi:
…ponse

fixes #695
Distance and duration might be 0 if the same coord is sent twice. They should be included nevertheless.

### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ RELEASING:
 - API endpoint "centrality" to calculate [betweenness centrality](https://en.wikipedia.org/wiki/Betweenness_centrality) values for nodes inside a given bounding box. Centrality is calculated using Brandes' algorithm. 
 ### Fixed
 - Fixed calculation of route distance limits with skipped segments ([#814](https://github.com/GIScience/openrouteservice/issues/814))
+- fixed missing segment distance and duration ([#695](https://github.com/GIScience/openrouteservice/issues/695)
 
 ## [6.3.6] - 2021-02-02
 ### Fixed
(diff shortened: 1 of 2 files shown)


takb (approved):
ok

PR #875, opened 2021-03-18
fix algo selection for continue_straight without bearings param

### LINKED ISSUE #702: Bearings property cannot be used on any profile
#### Here's what I did

Requested a route that made use of the bearings API parameter, i.e. 
```
https://api.openrouteservice.org/v2/directions/cycling-mountain

{
	"coordinates": [[8.681495,49.41461],[8.687872,49.420318]],
	"bearings": [[25,30],[90,20]]
}
```

---
#### Here's what I got

An error message saying that headings is not supported
```
{
    "error": {
        "code": 2099,
        "message": "Heading is not (fully) supported for CHGraph. See issue #483"
    },
    "info": {
        "engine": {
            "version": "6.1.0",
            "build_date": "2020-04-27T04:45:15Z"
        },
        "timestamp": 1588226836862
    }
}
```

---

That the routing process takes the preferred bearings into account as has previously been the case 

---

It seems there has been some change in the logic as previously when bearings was passed the system dropped out of using CH as bearings were not possible with that. Now this is not the case and when bearings are provided it does not fall back. This can be seen by adding the `"preference": "shortest"` parameter which then falls out of CH as shortest does not have the CH graph built

takb:
### Pull Request Checklist

Fixes #702 .

- Reason for change:
title says all


rabidllama (changes requested):
All looks good and works, but the API request documentation tags need updating if this is being allowed in the driving profiles aswell

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequest.java
```diff
@@ -178,8 +178,8 @@
     private boolean hasBearings = false;
 
     @ApiModelProperty(name = PARAM_CONTINUE_STRAIGHT,
-            value = "Forces the route to keep going straight at waypoints restricting uturns there even if it would be faster. This setting will work for all profiles except for `driving-*`. " +
-            "CUSTOM_KEYS:{'apiDefault':'true','validWhen':{'ref':'profile','valueNot':['driving-*']}}",
+            value = "Forces the route to keep going straight at waypoints restricting uturns there even if it would be faster." +
+            "CUSTOM_KEYS:{'apiDefault':'true'}",
```
rabidllama: I know this is from before, but is this true? I thought that by default continue straight was disabled as it would force the use of slower algorithms.

rabidllama (changes requested):

openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequest.java
takb: you're right, the actual default is false, corrected docs

rabidllama (approved):

PR #885, opened 2021-03-29
fix indexing for consecutive identical coordinates

### LINKED ISSUE #762: Invalid Way Point indexes returned in Directions V2 Service.
Hello, I'm currently building an app that uses the Directions V2 service, and I got an error while implementing it. The two way points returned for the steps that are nearly at the end of the route coordinates, sometimes are pointing to a position that don't actually exist in the coordinates array. This is not the first time that I'm running into this error, I have opened an issue with a similar error using the directions V1 service (https://github.com/GIScience/openrouteservice/issues/751), and I was told to use the V2 service, and clearly the error is still there.

#### Here's what I did
A POST request to this URL: https://api.openrouteservice.org/v2/directions/driving-car
With that JSON body: 
`{
	"attributes": [
		"avgspeed",
		"detourfactor",
		"percentage"
	],
	"coordinates": [
		[
			-43.167751,
			-19.83567
		],
		[
			-43.167751,
			-19.83567
		]
	],
	"elevation": true,
	"extra_info": [
		"surface"
	],
	"instructions_format": "html",
	"maneuvers": true
}`

---
#### Here's what I got
`{
  "routes": [
    {
      "summary": {},
      "segments": [
        {
          "steps": [
            {
              "distance": 0.0,
              "duration": 0.0,
              "type": 11,
              "instruction": "Head east",
              "name": "-",
              "way_points": [
                0,
                1
              ],
              "maneuver": {
                "bearing_before": 0,
                "bearing_after": 0
              }
            }
          ]
        }
      ],
      "bbox": [
        -43.16776,
        -19.835674,
        853.0,
        -43.16776,
        -19.835674,
        853.0
      ],
      "geometry": "|caxBnenfGgreD",
      "way_points": [
        0,
        0
      ],
      "extras": {
        "surface": {
          "values": [],
          "summary": []
        }
      }
    }
  ],
  "bbox": [
    -43.16776,
    -19.835674,
    853.0,
    -43.16776,
    -19.835674,
    853.0
  ],
  "metadata": {
    "attribution": "openrouteservice.org | OpenStreetMap contributors",
    "service": "routing",
    "timestamp": 1593664694848,
    "query": {
      "coordinates": [
        [
          -43.167751,
          -19.83567
        ],
        [
          -43.167751,
          -19.83567
        ]
      ],
      "profile": "driving-car",
      "format": "json",
      "instructions_format": "html",
      "attributes": [
        "avgspeed",
        "detourfactor",
        "percentage"
      ],
      "maneuvers": true,
      "elevation": true,
      "extra_info": [
        "surface"
      ]
    },
    "engine": {
      "version": "6.1.1",
      "build_date": "2020-06-18T15:14:48Z",
      "graph_date": "1970-01-01T00:00:00Z"
    }
  }
}`

As in the response, even though the segment's way point indexes are correct, the steps' point to coordinates that don't exist (index 1), since the encoded geometry array has only one position. Another request with the same JSON body can be made to https://api.openrouteservice.org/v2/directions/driving-car/geojson in order to visualize the coordinates as geojson, without encoding, and the same error still happens. In this example, the coordinates used to do the routing are the same, but using another sets of coordinates would probably have the chance of getting the same error too, since the error is similar to (https://github.com/GIScience/openrouteservice/issues/751), and probably the same bug is still present in the V2 version of the API.

---

No greater way point indexes than the coordinate array size - 1.

takb:
### Pull Request Checklist

Fixes #762 .

- Reason for change:
Fixed incorrect indexing of waypoints in route request responses due to consecutive identical coordinates.

### DIFF
--- a/openrouteservice/src/main/java/org/heigit/ors/routing/RouteResultBuilder.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/routing/RouteResultBuilder.java
@@ -322,7 +322,10 @@ class RouteResultBuilder
 	}
 
 	private int getEndWayPointIndex(int startIndex, InstructionType instrType, Instruction instr) {
-		if (instrType == InstructionType.FINISH || (instrType == InstructionType.DEPART && instr.getDistance() == 0.0 && instr.getPoints().size() == 1))
+		if (instrType == InstructionType.FINISH
+                // "empty" departure instruction means start and end coordinates are the same, index should not increase
+                || (instrType == InstructionType.DEPART && instr.getDistance() == 0.0 && instr.getPoints().size() == 1)
+            )
 			return startIndex;
 		else
 			return startIndex + instr.getPoints().size();
--- a/openrouteservice/src/main/java/org/heigit/ors/routing/RouteResultBuilder.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/routing/RouteResultBuilder.java
@@ -154,7 +154,8 @@ class RouteResultBuilder
     }
 
     private void setDepartureArrivalTimes(String timezoneDeparture, String timezoneArrival, RoutingRequest request, RouteResult result) {
-        ZonedDateTime departure, arrival;
+        ZonedDateTime departure;
+        ZonedDateTime arrival;
 
         long duration = (long) result.getSummary().getDuration();
 
@@ -321,7 +322,7 @@ class RouteResultBuilder
 	}
 
 	private int getEndWayPointIndex(int startIndex, InstructionType instrType, Instruction instr) {
-		if (instrType == InstructionType.FINISH)
+		if (instrType == InstructionType.FINISH || (instrType == InstructionType.DEPART && instr.getDistance() == 0.0 && instr.getPoints().size() == 1))
 			return startIndex;
 		else
 			return startIndex + instr.getPoints().size();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ RELEASING:
  -->
 
 ## [Unreleased]
+### Fixed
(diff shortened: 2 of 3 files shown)


openrouteservice/src/main/java/org/heigit/ors/routing/RouteResultBuilder.java
```diff
@@ -321,7 +322,7 @@ else if (sign == 1)
 	}
 
 	private int getEndWayPointIndex(int startIndex, InstructionType instrType, Instruction instr) {
-		if (instrType == InstructionType.FINISH)
+		if (instrType == InstructionType.FINISH || (instrType == InstructionType.DEPART && instr.getDistance() == 0.0 && instr.getPoints().size() == 1))
```
sfendrich: Can we make this more readable? I think it is not really clear, what this condition means and why it covers the issue to be fixed. For example, it is unintuitive that the conjunction contains `InstructionType.DEPART` although the unit test has identical coordinates in the middle.

takb: Yes, but the way GH returns points for many cases is unintuitive and all conditions are required to not let some other cases fail. I already added brackets that are not required to improve readability. All I could think of is add some inline docs...

takb: >  it is unintuitive that the conjunction contains `InstructionType.DEPART` although the unit test has identical coordinates in the middle.

and that's because ORS requests each segment of the route separately to GH, so two consecutive identical coordinates are handled as one query regardless where they occur in a list of coordinates.

sfendrich (approved):

PR #902, opened 2021-04-14
set HGV speed calculation parameter in app.config.sample

### LINKED ISSUE #806: API Playground Results vs Self Hosted very different (Durations)
#### Here's what I did
I spun up a linux VM as I needed to produce distance/duration metrics for around 65000 location pairs.  I could not run using the online version due to rate limits so spinning up my own version for this batch seemed like the easiest route.

I download the latest PBF file for Australia from http://download.geofabrik.de/australia-oceania/australia.html, spun up the docker image using the documentaion provided on your site and enabled the HGV profile.

---
#### Here's what I got
After running the batch, I noticed that the distances matches spot on to the online API playground, but the durations were very different. (much faster).

I checked a few samples using Google Maps too and Google Maps matches the API Playground durations.  Why would my self-hosted version produce much different duration results?

I have used the sample App.Config provided and made a single change to enable HGV profile.

**produced from my version**
`"durations": [
        [
            0.0,
            2809.81
        ],
        [
            2566.62,
            0.0
        ]
    ],
    "distances": [
        [
            0.0,
            51378.84
        ],
        [
            48858.77,
            0.0
        ]
    ],`

**produced from your API Playground**

`"durations": [
        [
            0.0,
            3426.16
        ],
        [
            3316.46,
            0.0
        ]
    ],
    "distances": [
        [
            0.0,
            51375.32
        ],
        [
            48847.5,
            0.0
        ]
    ]`

**request for both**
`{
  "locations": [
    [
      150.8144, -33.7861
    ],
    [
      150.7284, -34.0703
    ]
  ],
  "metrics": [
    "distance",
    "duration"
  ]
}`

takb:
### Pull Request Checklist

Fixes #806 .

- Reason for change:
app.config.sample HGV profile has now same settings regarding speed calculation as public API, to avoid user confusion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ RELEASING:
 ## [Unreleased]
 ### Added
 - Allow to disable OSM conditional access and speed encoders via parameter in config file
+### Changed
+- app.config.sample HGV profile has now same settings regarding speed calculation as public API ([#806](https://github.com/GIScience/openrouteservice/issues/806))
 ### Fixed
 - Fixed isochrones algorithm selection for location_type parameter ([#676](https://github.com/GIScience/openrouteservice/issues/676)
 - Updated link to client translations in readme
--- a/openrouteservice-api-tests/.gitignore
+++ b/openrouteservice-api-tests/.gitignore
@@ -1,4 +1,5 @@
 /target/
 /.classpath
 /.gitignore
-/data/elevation/srtm_38_03.gh
+/data/elevation/*
+!/data/elevation/srtm_38_03.zip
--- a/openrouteservice/src/main/resources/app.config.sample
+++ b/openrouteservice/src/main/resources/app.config.sample
@@ -177,7 +177,7 @@
             "profiles": "driving-hgv",
             "parameters": {
               "encoder_flags_size": 8,
-              "encoder_options": "turn_costs=true|block_fords=false|use_acceleration=false",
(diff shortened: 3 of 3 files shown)


aoles:
I don't immediately see how this is linked to #806. `maximum_speed_lower_bound` is only relevant for CALT preparation and can make a difference only for queries with `maximum_speed` set. The referenced issue never mentions this parameter.

takb:
Then I've got something mixed up while trying to reproduce the discrepancy, back to trials...

aoles:
`use_acceleration=false` is for 99,9% the reason for the observed discrepancies...

takb:
confirmed, thx - it got me confused because I'd set all parameters that were different from my local app.config, and it took almost an hour between each test run...

aoles (changes requested):
Thanks for cleaning-up the sample config file! I have the following improvement suggestions. 

First of all, I believe that if we enable acceleration for hgv, then we should do the same for the car profile as well. At least `use_acceleration=true` is what both profiles have set in production.

Secondly, I'm not sure why the hgv profile has both `lm` and `core` preparations enabled. I would keep only the latter one unless there are reasons to keep them both. 

Last but not least, `maximum_speed_lower_bound` should remain unset for all profiles.

Cheers,
Andrzej

aoles (approved):
Looks perfect now, thanks a million! :)

PR #929, opened 2021-05-06
Use as default fallback A* rather than Dijkstra

### LINKED ISSUE #926: Robustify default routing algorithm selection logic
Make sure A* with beeline approximation rather than plain bidir Dijkstra is used as the fallback routing algorithm. This was not always the case before https://github.com/GIScience/openrouteservice/pull/915/commits/7f762f7c5730bb623c121007173a38641a941c13#. For example, without ALT enabled Dijkstra was used when queried with avoid polygons.

aoles:
### Pull Request Checklist

Fixes #926 .

- Reason for change: There have been issues in the past that in certain constellations routing defaulted to Dijkstra rather than A*.

### DIFF
--- a/openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
@@ -111,7 +111,6 @@ public class RoutingProfile {
     private static final String KEY_ACTIVE_LANDMARKS = "active_landmarks";
     private static final String KEY_TOTAL_POP = "total_pop";
     private static final String KEY_TOTAL_AREA_KM = "total_area_km";
-    private static final String KEY_ASTARBI = Parameters.Algorithms.ASTAR_BI;
     private static final int KEY_FLEX_STATIC = 0;
     private static final int KEY_FLEX_PREPROCESSED = 1;
     private static final int KEY_FLEX_FULLY = 2;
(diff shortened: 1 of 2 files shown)


openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
```diff
@@ -1072,8 +1071,8 @@ private void setSpeedups(GHRequest req, boolean useCH, boolean useCore, boolean
         req.getHints().put(KEY_CORE_DISABLE, !useCore);
         req.getHints().put(KEY_LM_DISABLE, !useALT);
 
-        if(!useCH)
-            req.setAlgorithm(KEY_ASTARBI);
+        if (useCH)
+            req.setAlgorithm(Parameters.Algorithms.DIJKSTRA_BI);
```
takb, line 1071: So we still want to fall back to plain dijkstra if CH is enabled?

aoles: This should be fine, as CH uses Dijkstra rather than A*.  Once `useCH` has been set both CH is available for the given profile and weighting, and the request parameters are compatible with CH routing. So I don't think it can switch to a different algorithm afterwards.

takb: ok

takb (approved):
lgtm

PR #947, opened 2021-05-18
remove String.format for converting parameter

### LINKED ISSUE #946: Appling Weightings from request parameter fail on a german local enviroment.
Floating point weights are converted to `String` in the process of setting profile parameter.  For the conversion `String.format("%.2f",..)` is used which depends on the default `Local` of the environment.
e.g. on a German local environment the decimal separator is "," instead of "." and so the converted string of e.g. "`0.5`" becomes "`0,5`" which results in a "NumberFormatException" if later this string should be  converted back to a floating point!

Affected lines are:
- [GenericHandler#L319](https://github.com/GIScience/openrouteservice/blob/104371bb60298f90641908a1d11e3e6a6325eace/openrouteservice/src/main/java/org/heigit/ors/api/requests/common/GenericHandler.java#L319)
- [GenericHandler#L328](https://github.com/GIScience/openrouteservice/blob/104371bb60298f90641908a1d11e3e6a6325eace/openrouteservice/src/main/java/org/heigit/ors/api/requests/common/GenericHandler.java#L328)

rtroilo:
### Pull Request Checklist

- ~3. I have documented my code using JDocs tags.~
- ~4. I have removed unnecessary commented out code, imports and System.out.println statements.~
- ~6. I have created API tests for any new functionality exposed to the API.~
- ~7. If changes/additions are made to the app.config file, I have added these to the app.config wiki page on github along with a short description of what it is for, and documented this in the Pull Request (below).~
- ~8. I have built graphs with my code of the Heidelberg.osm.gz file and run the api-tests with all test passing~
- ~10. For new features or changes involving building of graphs, I have tested on a larger dataset (at least Germany) and the graphs build without problems (i.e. no out-of-memory errors).~
- ~11. For new features or changes involving the graphbuilding process (i.e. changing encoders, updating the importer etc.), I have generated longer distance routes for the affected profiles with different options (avoid features, max weight etc.) and compared these with the routes of the same parameters and start/end points generated from the current live ORS. If there are differences then the reasoning for these **MUST** be documented in the pull request.~

Fixes #946 .

- Reason for change: `String.format` for converting floating points without a given Local use the System default which could result in a fail to converted it back to floating points.

This PR removes the need to convert parameters with `String.format`. 

⚠️ THIS PR CHANGE SLIGHTLY THE LOGIC! Profile weights (greenFactor, quietFactor) were truncated to two decimal places only, this PR removes this truncation! ⚠️

### DIFF
--- a/openrouteservice/src/main/java/org/heigit/ors/api/requests/common/GenericHandler.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/api/requests/common/GenericHandler.java
@@ -41,6 +41,7 @@ import org.json.simple.JSONObject;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 public class GenericHandler {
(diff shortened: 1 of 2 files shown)


takb (approved):
looks good to me, thx!

PR #950, opened 2021-05-20
simplify radiuses usage

### LINKED ISSUE #923: Simplify usage of radiuses-Parameter
#### Here's what I did
I wanted to find a route using multiple waypoints and set the search radiuses for finding routable points higher than the default 350m, but the same for all my waypoints.

---
#### Here's what I got
I had to input individual radiuses for each point, which was tedious.

---
I was hoping for the openrouteservice to be a touch user-friendlier ;)

---
The `radiuses` parameter could be expanded to also take a single value, in which case this value would be used as the search radius for all specified points, or as a list as it currently is.

koebi:
### Pull Request Checklist

Fixes #923 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ RELEASING:
  -->
 
 ## [Unreleased]
+### Fixed
+- Accept single value and array of length 1 as `radiuses`-parameter ([#923](https://github.com/GIScience/openrouteservice/issues/923))
 
 ## [6.5.0] - 2021-05-17
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ RELEASING:
  -->
 
 ## [Unreleased]
-### Fixed
+### Added
 - Accept single value and array of length 1 as `radiuses`-parameter ([#923](https://github.com/GIScience/openrouteservice/issues/923))
 
 ## [6.5.0] - 2021-05-17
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ RELEASING:
  -->
 
 ## [Unreleased]
+### Fixed
```
TheGreatRefrigerator: New features should go under
`### Added`
as this is not a fix but an enhancement

e.g. 
```
### Added
- possibility to pass single `radiuses`-parameter value for all waypoints ( and the rest)
```

In this case you could also put under `### Changed`
as it changed the functionality of the `radiuses` param, but i think added fit's better, as all the previous functionality is still there.

koebi: In fact, it is not - where previously a request with more than one coordinate and only one radius value would yield a parameter value/format error, it now doesn't anymore.
The original commit's message states that as well - this is not backwards compatible ;)

TheGreatRefrigerator: > where previously a request with more than one coordinate and only one radius value would yield a parameter value/format error, it now doesn't anymore.

But i thought that is exactly the added feature, no?

Current documenation states:
radiuses

so an error should be expected if the radiuses count currently doesn't match the coordinate count.

also https://github.com/GIScience/openrouteservice/issues/923 is labeled as enhancement | feature

I don't see a reason to put it in `Fixed` category 🤷

takb: I agree with @TheGreatRefrigerator, the previous behavior was documented correctly so this is not a fix. Since this creates backwards-incompatibility this needs to go in a minor version up release.

takb (approved):
lgtm

PR #957, opened 2021-05-26
Do not use virtual edges for building isochrone geometry

HendrikLeuschner:
### Pull Request Checklist

- [x] 13. For changes touching the API documentation, i have tested that the API playground [renders correctly](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/openrouteservice-infrastructure/ors-docs-api#test-new-ors-documentation).

Fixes # .

- Key functionality added: Check whether edge is virtual before processing in isochrone geometry. Virtual edges have no geometry -> fails. Was added because one of the last changes added QueryGraph usage for isochrones for fixing another bug.
- Reason for change:

-

-

### DIFF
--- a/openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
@@ -332,6 +332,7 @@ public class ConcaveBallsIsochroneMapBuilder implements IsochroneMapBuilder {
 		GraphHopperStorage graph = searchContext.getGraphHopper().getGraphHopperStorage();
 		NodeAccess nodeAccess = graph.getNodeAccess();
 		int maxNodeId = graph.getNodes();
+		int maxEdgeId = graph.getEdges() - 1;
 
 		DistanceCalc dcFast = new DistancePlaneProjection();
 		double bufferSize = 0.0018;
@@ -365,7 +366,7 @@ public class ConcaveBallsIsochroneMapBuilder implements IsochroneMapBuilder {
 			edgeId = goalEdge.originalEdge;
 			nodeId = goalEdge.adjNode;
 
-			if (edgeId == -1 || nodeId == -1 || nodeId > maxNodeId)
+			if (edgeId == -1 || nodeId == -1 || nodeId > maxNodeId || edgeId > maxEdgeId)
 				continue;
 
 			float maxCost = (float) goalEdge.weight;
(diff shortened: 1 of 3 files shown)


openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
```diff
@@ -332,6 +332,7 @@ private GeometryCollection buildIsochrone(AccessibilityMap edgeMap, List<Coordin
 		GraphHopperStorage graph = searchContext.getGraphHopper().getGraphHopperStorage();
 		NodeAccess nodeAccess = graph.getNodeAccess();
 		int maxNodeId = graph.getNodes();
```
aoles: In node indexing is 0-based and `getNodes()` returns the number of nodes, then shouldn't this rather be `int maxNodeId = graph.getNodes() - 1;`?

aoles (changes requested):

openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
HendrikLeuschner: Citing GH 
```The virtual nodes and edges have a higher int ID than graph.getNodes() or allEdges.length()``` https://github.com/graphhopper/graphhopper/blob/master/docs/core/low-level-api.md
This is at least for the edges not true. There it really is the count. The node part was already in and I did not change or check it.

aoles: Apparently that's not entirely true according to https://github.com/GIScience/graphhopper/blob/ors_0.13/core/src/main/java/com/graphhopper/routing/QueryGraph.java#L375-L381

HendrikLeuschner: Yes, as I said, for the edges it is not true (I use id > maxId -1, QueryGraph uses id >= maxId). So seems that its also not true for the nodes, if we can trust QueryGraph impl. So generally the GH explanation is wrong.

HendrikLeuschner: Changed it

aoles: I believe both `isVirtualEdge` and `isVirtualNode` methods are correct considering other bits of GH code I've seen such as https://github.com/GIScience/graphhopper/blob/ors_0.13/core/src/main/java/com/graphhopper/storage/BaseGraph.java#L69-L72 and https://github.com/GIScience/graphhopper/blob/ors_0.13/core/src/main/java/com/graphhopper/storage/BaseGraph.java#L86-89, respectively.

aoles (approved):
LGTM, thanks for fixing this!

Issue #956, opened 2021-05-26
Enable time-dependent speeds for any weighting


aoles:
Time-dependent speed needs to be taken into account not only for weightings which depend on speed such as `fastest` but for any weighting in order to compute edge travel time. Currently time-dependent speed is supported only by `TimeDependentFastestWeighting`. In particular, `OptimizedPriorityWeighting` and `ShortestWeighting` do not consider time dependent speed when calculating route duration which leads to [wrong results](https://github.com/GIScience/openrouteservice/blob/master/openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/routing/ResultTest.java#L3373).

PR #960, opened 2021-05-27
Refactor weightings to use speed calculator

### LINKED ISSUE #955: Maximum speed limit not compatible with priority type weightings
#### Here's what I did
Queried for a recommended HGV route with and without maximum speed value set.

---
#### Here's what I got
Default route without any additional settings:
https://classic-maps.openrouteservice.org/directions?n1=48.276382&n2=11.577632&n3=15&a=48.259512,11.557102,48.292616,11.573326&b=4a&c=0&k1=en-US&k2=km
image

Route with maximum speed set to 84 km/h:
https://classic-maps.openrouteservice.org/directions?n1=48.276382&n2=11.577632&n3=15&a=48.259512,11.557102,48.292616,11.573326&b=4a&c=0&d=84&k1=en-US&k2=km
image

---

No change of route for `maximumn_speed = 84` in this particular example.

---

The current implementation of `MaximumSpeedWeighting` is not compatible with `OptimizedPriorityWeighting` because if for a given edge the maximum speed is used then its weight is not scaled by priority.

aoles:
### Pull Request Checklist

**IMPORTANT**: This PR is set-up to run against GH v0.13.14 which is expected to include https://github.com/GIScience/graphhopper/pull/40.

Fixes #955 and #956.

Introduces the concept of speed calculators for computing travel time across weightings in order to allow combining different scenarios involving variable edge speeds:
- taking into account traffic data
- OSM time-dependent conditional speed limits
- user adjustable maximum speed


openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/util/ORSParameters.java
```diff
@@ -25,6 +25,14 @@ private ORSParameters() {}
     /* Parameters with an 'INIT' prefix are used as defaults and/or are configured at start.*/
     static final String ROUTING_INIT_PREFIX = "routing.";
 
+    /**
+     * Parameters that can be passed as hints and influence routing per request.
+     */
+    public static final class Routing {
```
HendrikLeuschner: I find `Routing` as class name a bit confusing here. This should be usable for matrix, isochrones too, correct? Maybe something like `Weighting`? Or `Calculation`? Idk

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/OptimizedPriorityWeighting.java
```diff
@@ -34,8 +34,8 @@ public OptimizedPriorityWeighting(FlagEncoder encoder, PMap map) {
 	}
 
 	@Override
-	public double calcWeight(EdgeIteratorState edgeState, boolean reverse, int prevOrNextEdgeId) {
-		double weight = super.calcWeight(edgeState, reverse, prevOrNextEdgeId);
+	public double calcWeight(EdgeIteratorState edgeState, boolean reverse, int prevOrNextEdgeId, long edgeEnterTime) {
+		double weight = super.calcWeight(edgeState, reverse, prevOrNextEdgeId, edgeEnterTime);
```
HendrikLeuschner, line 38: What's with all the other weightings in `routing/graphhopper/extensions/weighting`? Do they not need this adaptation?

aoles: Good point! This depends on whether a given weighting calls `super.calcWeight` and/or whether it depends on the edge speed. I will go over individual weightings and make adjustments if necessary.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/util/ORSParameters.java
aoles: Thanks, done!

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/OptimizedPriorityWeighting.java
aoles: I've only changed `AdditionWeighting` and `FastestSafeWeighting`. `AdditionWeighting` is used to collate the so-called soft weightings such as `AccelerationWeighting`, `AvoidHillsWeighting`, `GreenWeighting`, `QuietWeighting` and `SteepnessDifficultyWeighting`. These are not "true" standalone weightings but rather provide a modifying factor applied to the base weighting. As such they do not need any adaptations because for the time being they are not time dependent, nor they rely on the speed calculator pipeline. `FastestSafeWeighting` is not currently being used in our setup, yet it is the weighting to which `ORSWeightingFactory` will default if no `shortest`, `fastest` or `recommended` weighting is specified and the flag encoder supports `PriorityWeighting`. 

Furthermore, no changes are necessary to `PreferencePriorityWeighting` or `ConstantWeighting` as the former one doesn't seem to be used in any scenario, while the latter one is defined just for the sake  of unit tests.

Last but not least, I've left `DistanceWeighting` as is. It should be fine for now, nevertheless in long run we might want to look into replacing it by regular `shortest` weighting which it effectively seems to be apart from additionally checking for `speed = 0`. However, this check is probably obsolete because of https://github.com/GIScience/openrouteservice/pull/945. I will open a separate issue for it later.

HendrikLeuschner (approved):

PR #992, opened 2021-07-01
Traffic-enabled routing and isochrones

aoles:
### Pull Request Checklist

- [ ] 13. For changes touching the API documentation, i have tested that the API playground [renders correctly](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/openrouteservice-infrastructure/ors-docs-api#test-new-ors-documentation).

Fixes # .

- Key functionality added:
- Reason for change:

-

-


openrouteservice/src/main/java/org/heigit/ors/routing/traffic/TrafficSpeedCalculator.java
```diff
@@ -0,0 +1,87 @@
...
+import java.time.ZonedDateTime;
+
+public class TrafficSpeedCalculator extends AbstractAdjustedSpeedCalculator {
+    // time-dependent stuff
+    protected TrafficGraphStorage trafficGraphStorage;
+    protected int timeZoneOffset;
+    private VehicleFlagEncoder vehicleFlagEncoder;
+    private boolean isVehicle = false;
+    private boolean isHGV = false;
+    private double HGVTrafficSpeedLimit = 80.0;
```
aoles, line 24: @HendrikLeuschner have you seen `RoutingTrafficSpeedCalculator` which is used for traffic-enabled routing? It wraps around `TrafficSpeedCalculator` and uses a similar approach to yours of limiting the speed up to a certain value. The main difference is that the limit is based on the native flag encoder field `maxPossibleSpeed` rather than a dedicated value `HGVTrafficSpeedLimit`. `maxPossibleSpeed` corresponds to the upper bound of speed values stored by the given flag encoder. The current limits for car and hgv are set to 140 and 90 km/h, respectively. The choice of the hgv value is based on the maximum value of country-specific speed limits defined in [heavyvehicle.json](https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/resources/resources/services/routing/speed_limits/heavyvehicle.json).

The solution based on  `maxPossibleSpeed` might have the advantage of being more scalable as it works for any given profile and doesn't require the additional logic to check for the encoder instance. In long run it's probably easier to maintain as well. What do you think? I'm asking because we might want to unify the speed calculation pipeline such that both isochrone and direction endpoints use one and the same speed calculator.

Looking forward to your feedback. Cheers!

HendrikLeuschner: Hi Andrzej. Yes I have seen that.
Initially, I added that same logic, too. The problem turned out to be that `flagEncoder.getMaxSpeed()` returned `0.0`. I do not know why, but this is breaking. Probably something in my config.
Secondly, this is not exactly what the requirements from the customer request: (I am aware of the fact that the following requirement is not sensible, but I am not the one making the requirements)
The customer explicitly requested that _traffic speeds_ are only considered for HGV up to `80kmh`. Now HGV do not only go `80kmh`, but usually `100kmh`, at least in DE. This will lead to the scenario where cars go `81` due to traffic, but hgv can go e.g. `100`, because it is the speed from the flagencoder. This is _not_ the same as the solution you have proposed, which would limit the speed entirely.

aoles: Thanks Hendrik for getting back to me! 

> The customer explicitly requested that _traffic speeds_ are only considered for HGV up to `80kmh`. Now HGV do not only go `80kmh`, but usually `100kmh`, at least in DE. This will lead to the scenario where cars go `81` due to traffic, but hgv can go e.g. `100`, because it is the speed from the flagencoder.

So if understand it correctly for hgv any traffic speed higher than 80 km/h will result in using the default encoder value even if it is higher than the traffic speed? If indeed this is the case then to me such logic doesn't make much sense either 🤷‍♂️  Though for Germany the only roads for which hgv flag encoder uses values higher than 80 km/h are motorways. For these the speed is set to 85 km/h by [heavyvehicle.json](https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/resources/resources/services/routing/speed_limits/heavyvehicle.json#L4) but only if the given road section is not tagged with `zone:traffic=DE:motorway` because then it falls back to 80 km/h again.

HendrikLeuschner: you got it

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
```diff
@@ -1093,9 +1074,12 @@ private void setSpeedups(GHRequest req, boolean useCH, boolean useCore, boolean
...
 
     boolean hasTimeDependentSpeed (RouteSearchParameters searchParams, RouteSearchContext searchCntx) {
+        if (!searchParams.isTimeDependent())
+            return false;
+
         FlagEncoder flagEncoder = searchCntx.getEncoder();
-        String key = EncodingManager.getKey(flagEncoder, ConditionalEdges.SPEED);
-        return searchParams.isTimeDependent() && flagEncoder.hasEncodedValue(key);
+        return flagEncoder.hasEncodedValue(EncodingManager.getKey(flagEncoder, ConditionalEdges.SPEED))
+                || mGraphHopper.isTrafficEnabled();
```
HendrikLeuschner: This makes the weighting time dependent independently of whether the request actually wants time-dependence. Will this not make every request time dependent if you have any traffic data? Seems wrong to me.

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
```diff
@@ -1105,7 +1089,8 @@ boolean requiresTimeDependentWeighting(RouteSearchParameters searchParams, Route
         FlagEncoder flagEncoder = searchCntx.getEncoder();
 
         return flagEncoder.hasEncodedValue(EncodingManager.getKey(flagEncoder, ConditionalEdges.ACCESS))
-                || flagEncoder.hasEncodedValue(EncodingManager.getKey(flagEncoder, ConditionalEdges.SPEED));
+                || flagEncoder.hasEncodedValue(EncodingManager.getKey(flagEncoder, ConditionalEdges.SPEED))
+                || mGraphHopper.isTrafficEnabled();
```
HendrikLeuschner, line 1061: Potentially same as above?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/GraphProcessContext.java
```diff
@@ -121,13 +122,17 @@ public void processWay(ReaderWay way)  {
...
 	 */
-	public void processWay(ReaderWay way, Coordinate[] coords, HashMap<Integer, HashMap<String, String>> nodeTags) {
+	public void processWay(ReaderWay way, Coordinate[] coords, HashMap<Integer, HashMap<String, String>> nodeTags, Coordinate[] allCoordinates) {
 		try {
 			if (arrStorageBuilders != null) {
 				int nStorages = arrStorageBuilders.length;
 				if (nStorages > 0) {
 					for (int i = 0; i < nStorages; ++i) {
-						arrStorageBuilders[i].processWay(way, coords, nodeTags);
+						if (arrStorageBuilders[i].getName().equals(HereTrafficGraphStorageBuilder.BUILDER_NAME)){
```
HendrikLeuschner: This could be more efficiently solved by storing which one of the builders from the array is the `HereTrafficGraphStorageBuilder`. Given that this function is called about a billion times, might be worth changing.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReader.java
```diff
@@ -225,11 +233,18 @@ public void onProcessWay(ReaderWay way) {
 				for (int i=0; i<osmNodeIds.size(); i++) {
 					int id = getNodeMap().get(osmNodeIds.get(i));
 					try {
-						double lat = getLatitudeOfNode(id, true);
-						double lon = getLongitudeOfNode(id, true);
+						double lat = getLatitudeOfNode(id, false);
+						double lon = getLongitudeOfNode(id, false);
+						boolean condition = !(lat == 0 || lon == 0 || Double.isNaN(lat) || Double.isNaN(lon));
+						if (processWholeGeom && condition) {
+							// TODO check why no tower nodes were used before
```
HendrikLeuschner: There are two TODOs in this file. Please do them or remove.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReader.java
```diff
@@ -225,11 +233,18 @@ public void onProcessWay(ReaderWay way) {
 				for (int i=0; i<osmNodeIds.size(); i++) {
 					int id = getNodeMap().get(osmNodeIds.get(i));
 					try {
-						double lat = getLatitudeOfNode(id, true);
-						double lon = getLongitudeOfNode(id, true);
+						double lat = getLatitudeOfNode(id, false);
+						double lon = getLongitudeOfNode(id, false);
+						boolean condition = !(lat == 0 || lon == 0 || Double.isNaN(lat) || Double.isNaN(lon));
```
HendrikLeuschner: Please rename `condition` to something understandable

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/TrafficRelevantWayType.java
```diff
@@ -0,0 +1,64 @@
...
+    public static byte getHereTrafficClassFromOSMRoadType(short roadType) {
+        if (roadType == TrafficGraphStorage.MOTORWAY || roadType == TrafficGraphStorage.MOTORROAD) {
+            return TrafficRelevantWayType.CLASS1;
+        } else if (roadType == TrafficGraphStorage.PRIMARY || roadType == TrafficGraphStorage.TRUNK) {
+            return TrafficRelevantWayType.CLASS2;
+        } else if (roadType == TrafficGraphStorage.SECONDARY) {
+            return TrafficRelevantWayType.CLASS3;
+        } else if (roadType == TrafficGraphStorage.TERTIARY) {
+            return TrafficRelevantWayType.CLASS4;
+//        } else if (roadType == TrafficGraphStorage.UNCLASSIFIED) {
```
HendrikLeuschner: remove

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/CoreAlgoFactoryDecorator.java
```diff
@@ -382,6 +382,9 @@ private EdgeFilter createCoreEdgeFilter(CHProfile chProfile, GraphHopperStorage
             edgeFilterSequence.add(new TimeDependentCoreEdgeFilter(gs));
         }
 
+        //FIXME fail-save init
```
HendrikLeuschner: fixed?

HendrikLeuschner: Should really just not be added if there is no traffic graph storage

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/flagencoders/HeavyVehicleFlagEncoder.java
```diff
@@ -68,6 +68,8 @@ public HeavyVehicleFlagEncoder(PMap properties) {
     public HeavyVehicleFlagEncoder(int speedBits, double speedFactor, int maxTurnCosts) {
         super(speedBits, speedFactor, maxTurnCosts);
 
+        maxPossibleSpeed = 90;
```
HendrikLeuschner, line 71: I don't think this is ever used. Sets the field in `AbstractFlagEncoder` but the only use of that field is for the bike encoder. ?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/TrafficGraphStorage.java
```diff
@@ -0,0 +1,651 @@
...
+            return;
+
+        if (getEdgeIdTrafficPatternLookup(edgeId, baseNode, adjNode, weekday) > 0 && lastPriority > priority)
+            return;
+
+        long edgePointer = (long) edgeId * edgeLinkLookupEntryBytes;
+
+        priorityValue[0] = (byte) priority;
+
+        // TODO RAD
```
HendrikLeuschner: Please remove all unnecessary comments here

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/AbstractGraphStorageBuilder.java
```diff
@@ -32,6 +33,8 @@ public void setParameters(Map<String, String> parameters)
 	{
 		this.parameters = parameters;
 	}
-	
+
 	public void finish(){}
+
+    public void postProcess(ORSGraphHopper graphHopper) {}
```
HendrikLeuschner: Given the fact that this method is in the interface for a lot of classes but only ever implemented and used in the Here Storage, it would be better to simply type cast the storage in line 610 of `ORSGraphhopper`. Then you'd only have the postprocess method in the relevant storage and not cluttering anything else.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/GraphStorageBuilder.java
```diff
@@ -39,4 +40,6 @@
 	String getName();
 	
 	void finish();
+
+	void postProcess(ORSGraphHopper graphHopper);
```
HendrikLeuschner: see previous comment

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/HereTrafficGraphStorageBuilder.java
```diff
@@ -0,0 +1,452 @@
...
+
+    @Override
+    public void processEdge(ReaderWay way, EdgeIteratorState edge) {
+        // processEdge(ReaderWay way, EdgeIteratorState edge, com.vividsolutions.jts.geom.Coordinate[] coords) overwrites this function.
+        // If the coords are directly delivered it becomes much faster than querying it from the edge
+    }
+
+    @Override
+    public void processEdge(ReaderWay way, EdgeIteratorState edge, com.vividsolutions.jts.geom.Coordinate[] coords) {
+//        if (outputLog) {
```
HendrikLeuschner: remove

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/HereTrafficGraphStorageBuilder.java
```diff
@@ -0,0 +1,452 @@
...
+                LOGGER.error("Error creating MultiLineString Type. This should not happen!", e);
+            }
+            File osmFile = null;
+            File osmMatchedFile = null;
+            File hereMatchedFile = null;
+            File hereFile = null;
+            int decimals = 14;
+            GeometryJSON gjson = new GeometryJSON(decimals);
+            FeatureJSON featureJSON = new FeatureJSON(gjson);
+//            osmFile = new File(dateFormat.format(date) + "_radius_" + matchingRadius + "_OSM_edges_output.geojson");
```
HendrikLeuschner: Please remove all dev comments

openrouteservice/src/main/java/org/heigit/ors/util/CSVUtility.java
```diff
@@ -66,6 +66,37 @@ private CSVUtility() {}
...
+            while ((ln = br.readLine()) != null) {
+                if(ignoreHeader && lines.isEmpty() && !headerRead) {
+                    headerRead=true;
+                    continue;
+                }
+
+                lines.add(parseLine(ln,customDelimiter,customQuoteChar));
+            }
+        } catch (IOException ioe) {
+            LOGGER.error("Could not read from file: " + file);
```
HendrikLeuschner: It seems to me that in case the file could not be read, the whole process should be aborted instead of returning an empty collection. Otherwise the following code might continue to run on nothing and produce results that, while they might not be wrong, might make the user think that file readings were fine. Is that a valid assumption?

HendrikLeuschner (changes requested):

openrouteservice/src/main/java/org/heigit/ors/util/CSVUtility.java
MichaelsJP: Removed as anyway not used.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/HereTrafficGraphStorageBuilder.java
MichaelsJP: Removed

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/HereTrafficGraphStorageBuilder.java
MichaelsJP: Removed

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReader.java
MichaelsJP: Done

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReader.java
MichaelsJP: Done

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/GraphProcessContext.java
MichaelsJP: @HendrikLeuschner Saving the location now.

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
MichaelsJP: @aoles

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
MichaelsJP: @aoles

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/TrafficRelevantWayType.java
MichaelsJP: Removed

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/CoreAlgoFactoryDecorator.java
MichaelsJP: @aoles

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/flagencoders/HeavyVehicleFlagEncoder.java
MichaelsJP: @aoles

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/TrafficGraphStorage.java
MichaelsJP: Removed

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/GraphStorageBuilder.java
MichaelsJP: Done

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/AbstractGraphStorageBuilder.java
MichaelsJP: Done

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
aoles: You are not entirely right @HendrikLeuschner . Whether the request is time dependent or not is checked in the preceding line 1053, so the logic seems fine to me.

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
aoles: Same here @HendrikLeuschner, the time-dependence of the request is examined at the very beginning of the method.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/flagencoders/HeavyVehicleFlagEncoder.java
aoles: It is actually used by `RoutingTrafficSpeedCalculator` to set the corresponding field there through a call `maxPossibleSpeed = flagEncoder.getMaxSpeed();`.

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/isochrones/ResultTest.java
```diff
@@ -424,7 +424,7 @@ public void testSmoothingFactor() {
                 .then()
                 .body("any { it.key == 'type' }", is(true))
                 .body("any { it.key == 'features' }", is(true))
-                .body("features[0].geometry.coordinates[0].size", is(51))
+                .body("features[0].geometry.coordinates[0].size", is(49))
```
sfendrich: What is this test supposed to test? If the comment on lines 414-415 is true, then it only tests implementation details and not a requirement. This test should either test something useful or be removed.

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
```diff
@@ -20,57 +20,90 @@
...
+            //If changed, this needs to be adapted in the traffic storage, too
+            ZonedDateTime zdt = parameters.getRouteParameters().getDeparture().atZone(ZoneId.of("Europe/Berlin"));
+            trafficSpeedCalculator.setZonedDateTime(zdt);
+            int toId = parameters.getReverseDirection() ? fromId : Integer.MIN_VALUE;
+            fromId = parameters.getReverseDirection() ? Integer.MIN_VALUE : fromId;
+            tdDijkstraCostCondition.calcPath(fromId, toId, zdt.toInstant().toEpochMilli());
+            IntObjectMap<SPTEntry> edgeMap = tdDijkstraCostCondition.getMap();
+            return new AccessibilityMap(edgeMap, tdDijkstraCostCondition.getCurrentEdge(), snappedPosition);
+        } else {
+            // IMPORTANT: It only works with TraversalMode.NODE_BASED.
```
sfendrich, line 75: Please use an assertion instead of a comment.

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
```diff
@@ -20,57 +20,90 @@
...
+            TrafficSpeedCalculator trafficSpeedCalculator = new TrafficSpeedCalculator(weighting.getSpeedCalculator());
+            trafficSpeedCalculator.init(graph, encoder);
+            weighting.setSpeedCalculator(trafficSpeedCalculator);
+            if (HelperORS.getTurnCostExtensions(graph.getExtension()) != null)
+                weighting = new TurnWeighting(weighting, HelperORS.getTurnCostExtensions(graph.getExtension()));
+            TDDijkstraCostCondition tdDijkstraCostCondition = new TDDijkstraCostCondition(queryGraph, weighting, parameters.getMaximumRange(), parameters.getReverseDirection(),
+                    TraversalMode.NODE_BASED);
+            tdDijkstraCostCondition.setEdgeFilter(edgeFilter);
+            //Time is defined to be in UTC + 1 because original implementation was for German traffic data
+            //If changed, this needs to be adapted in the traffic storage, too
```
sfendrich: Such an implicit dependency should not be hidden in a comment. Can we make this configurable?

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
```diff
@@ -20,57 +20,90 @@
...
 
-		IntObjectMap<SPTEntry> edgeMap = dijkstraAlg.getMap();
-       return new AccessibilityMap(edgeMap, dijkstraAlg.getCurrentEdge(), snappedPosition);
-	}
+        if (fromId == -1)
+            throw new InternalServerException(IsochronesErrorCodes.UNKNOWN, "The closest node is null.");
+        Weighting weighting = createWeighting(parameters, encoder);
+
+        if (parameters.isTimeDependent()) {
+            //Time-dependent means traffic dependent for isochrones (for now)
```
sfendrich: The following is quite heavy code to put in an if branch. I would prefer, if it was moved into a private method with a well-chosen name.

openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
```diff
@@ -32,473 +33,595 @@
...
-		double meanSpeed = maxSpeed;
+
+    private RouteSearchContext searchContext;
+
+    public void initialize(RouteSearchContext searchContext) {
+        geometryFactory = new GeometryFactory();
+        this.searchContext = searchContext;
+    }
+
+    public IsochroneMap compute(IsochroneSearchParameters parameters) throws Exception {
```
sfendrich, line 66: Please refactor this method into several smaller ones.

openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
```diff
@@ -32,473 +33,595 @@
...
+        double maxLength = (intervalDegrees / 100f) * smoothingFactor;
 
-		treeSet = new TreeSet<>();
+        if (maxLength < MINIMUM_DISTANCE)
+            maxLength = MINIMUM_DISTANCE;
+        return maxLength;
+    }
 
-		List<Coordinate> isoPoints = new ArrayList<>((int)(1.2*edgeMap.getMap().size()));
+    private void addIsochrone(IsochroneMap isochroneMap, Coordinate[] points, double isoValue, double maxRadius, double meanRadius, float smoothingFactor) {
```
sfendrich, line 218: Please refactor this method into several smaller ones.

openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
```diff
@@ -32,473 +33,595 @@
...
-			double meanRadius = 0;
-			if (isochroneType == TravelRangeType.DISTANCE) {
-				maxRadius = isoValue;
-				meanRadius = isoValue;
-			} else {
-				maxRadius = metersPerSecond * isoValue;
-				meanRadius = meanMetersPerSecond * isoValue;
-				isochronesDifference = metersPerSecond * isochronesDifference;
-			}
+    private Coordinate[] buildIsochrone(AccessibilityMap edgeMap, List<Coordinate> points, double lon, double lat,
```
sfendrich, line 340: Please refactor this method into several smaller ones.

openrouteservice/src/main/java/org/heigit/ors/mapmatching/hmm/HiddenMarkovMapMatcher.java
```diff
@@ -25,353 +25,359 @@
...
+        graphHopper = gh;
+
+        encoder = gh.getEncodingManager().fetchEdgeEncoders().get(0);
+        GraphHopperStorage graph = gh.getGraphHopperStorage();
+        locationIndex = new LocationIndexMatch(graph,
+                (com.graphhopper.storage.index.LocationIndexTree) gh.getLocationIndex(), (int) searchRadius);
+    }
+
+    @Override
+    public RouteSegmentInfo[] match(Coordinate[] locations, boolean bothDirections) {
```
sfendrich, line 79: Please refactor this long method into several smaller ones.

openrouteservice/src/main/java/org/heigit/ors/mapmatching/hmm/HiddenMarkovMapMatcher.java
```diff
@@ -25,353 +25,359 @@
...
+        if (seg1 != null) {
+            result[0] = seg1;
+            result[1] = seg2;
+        } else {
+            result[0] = seg2;
+        }
+        return result;
+    }
+
+    private RouteSegmentInfo findRouteSegments(Coordinate[] z, MatchPoint[][] x, int nR, int nZ, double[] startProbs, double[][] emissionProbs, double[][] transProbs) {
```
sfendrich, line 172: Please refactor this long method into several smaller ones.

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
```diff
@@ -749,9 +739,9 @@ private RouteSearchContext createSearchContext(RouteSearchParameters searchParam
         props.putObj("routing_profile_params", profileParams);
 
         /*
-        * PARAMETERS FOR EdgeFilterFactory
-        * ======================================================================================================
-        */
+         * PARAMETERS FOR EdgeFilterFactory
```
sfendrich, line 742: Use single-line comments (`//`)

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
```diff
@@ -749,9 +739,9 @@ private RouteSearchContext createSearchContext(RouteSearchParameters searchParam
...
+         * ======================================================================================================
+         */
 
         /* Avoid areas */
```
sfendrich, line 746: This comment is useless, please remove.

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
```diff
@@ -777,9 +767,9 @@ else if (profileType == RoutingProfileType.WHEELCHAIR) {
 
         /* Avoid borders of some form */
```
sfendrich, line 768: This comment is useless, please remove.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/TrafficRelevantWayType.java
```diff
@@ -0,0 +1,64 @@
...
+    public static final byte CLASS1LINK = 7;
+    public static final byte CLASS2LINK = 8;
+    public static final byte CLASS3LINK = 9;
+    public static final byte CLASS4LINK = 10;
+    public static final byte UNWANTED = 0; // Not needed! Streets Here doesn't provide traffic data for anyways.
+    // Class 1 = motorway | motorroad
+    // Class 2 = motorway | trunk | primary
+    // Class 3 = primary | secondary
+    // Class 4 = secondary | tertiary
+    // Class 5 = residential | living_street | service | ...
```
sfendrich: Why are these comments repreated here?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/TrafficRelevantWayType.java
```diff
@@ -0,0 +1,64 @@
...
+        } else if (roadType == TrafficGraphStorage.PRIMARY || roadType == TrafficGraphStorage.TRUNK) {
+            return TrafficRelevantWayType.CLASS2;
+        } else if (roadType == TrafficGraphStorage.SECONDARY) {
+            return TrafficRelevantWayType.CLASS3;
+        } else if (roadType == TrafficGraphStorage.TERTIARY) {
+            return TrafficRelevantWayType.CLASS4;
+//        } else if (roadType == TrafficGraphStorage.UNCLASSIFIED) {
+//            return TrafficRelevantWayType.UNCLASSIFIED;
+//        } else if (roadType == TrafficGraphStorage.RESIDENTIAL) {
+//            return TrafficRelevantWayType.CLASS5;
```
sfendrich: Please remove commented-out code.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/reader/traffic/HereTrafficReader.java
```diff
@@ -0,0 +1,246 @@
...
+            throw ioe;
+        }
+        currentInstance = this;
+    }
+
+    public boolean isInitialized() {
+        return this.isInitialized;
+    }
+
+    private void generatePatterns(HashMap<Integer, EnumMap<TrafficEnums.TravelDirection, Integer[]>> referencePatterns, Map<Integer, TrafficPattern> patterns) {
```
sfendrich, line 110: Please apply clean-coding practices here.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/reader/traffic/TrafficEnums.java
```diff
@@ -0,0 +1,188 @@
+package org.heigit.ors.routing.graphhopper.extensions.reader.traffic;
+
+public class TrafficEnums {
```
sfendrich, line 3: Why is this a class and not a package?
Also `TrafficEnum` is a very generic name, considering that all the enums seem to be very specific to a certain data format.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/reader/traffic/TrafficLinkMetadata.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public TrafficEnums.FunctionalClass functionalClass() {
+        return functionalClass;
+    }
+
+    public int getFunctionalClassWithRamp() {
+        if (this.ramp == TrafficEnums.NoYesEnum.NO) {
+            return this.functionalClass.getFunctionalClass();
+        } else if (this.functionalClass == TrafficEnums.FunctionalClass.CLASS1) {
```
sfendrich, line 76: Use switch in stead of an if-else-chain to make the code more readable.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/TrafficGraphStorage.java
```diff
@@ -0,0 +1,618 @@
...
+import org.heigit.ors.routing.graphhopper.extensions.reader.traffic.TrafficEnums;
+
+import java.util.Calendar;
+import java.util.TimeZone;
+
+/**
+ * Graph storage class for the Border Restriction routing
+ */
+public class TrafficGraphStorage implements GraphExtension {
+
```
sfendrich, line 31: Why not use enums here?

openrouteservice/src/main/java/org/os_concavehull/ConcaveHull.java
```diff
@@ -0,0 +1,1037 @@
...
+		}
+	}
+
+	/** breadth first digging, supports multi parts
+	 * @param triChecker triangle checker to be used
+	 * @param allowMultiParts if multiple parts are to be generated
+	 * @param keepLineSeg if degenerated line segments should be kept
+	 * @return a collection of geometry that form the concave hull of the input data (may contains linestring as degenerated segments)
+	 */
+	public Collection<Geometry> getConcaveHullBFS(TriangleChecker triChecker, boolean allowMultiParts, boolean keepLineSeg) {
```
sfendrich: PLease refactor this method into several shorter ones.

openrouteservice/src/main/java/org/os_concavehull/ConcaveHull.java
```diff
@@ -0,0 +1,1037 @@
...
+	}    //
+
+	/** Experimental Concave hull construction with an extra metric to control the order of boundary edge "digging"
+	 * @param triChecker
+	 * @param m TriangleMetric for controlling the order of digging (default is the length of the boundary edge)
+	 * @param allowMultiParts
+	 * @param keepLineSeg
+	 * @return
+	 */
+	public Collection<Geometry> getConcaveHullMetric(TriangleChecker triChecker, TriangleMetric m, boolean allowMultiParts, boolean keepLineSeg) {
```
sfendrich: Please refactor this method into several shorter ones.

openrouteservice/src/main/java/org/os_concavehull/TriCheckerAlpha.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ *     http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * 
+ */
```
sfendrich: Why do we copy-pase foreign code instead of using it through a library?

openrouteservice/src/test/java/org/heigit/ors/routing/traffic/TrafficSpeedCalculatorTest.java
```diff
@@ -0,0 +1,121 @@
...
+        carEncoder.getAverageSpeedEnc().setDecimal(false, edgeFlags, originalEdgeSpeed);
+        EdgeIteratorState edgeIteratorState = createMockedEdgeIteratorState(10, edgeFlags, 0, 1, edgeId, 2, 3);
+        assertEquals(65, trafficSpeedCalculator.getSpeed(edgeIteratorState, false, 1), 1e-8);
+    }
+
+    private class MockTrafficStorage extends TrafficGraphStorage {
+
+        @Override
+        public int getSpeedValue(int edgeId, int baseNode, int adjNode, long unixMilliSeconds, int timeZoneOffset) {
+            if (edgeId == 1)
```
sfendrich: Replacing this if-chain with switch would make the code more readable and more performant.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/core/CoreAlgoFactoryDecorator.java
aoles: Done!

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
HendrikLeuschner: Okay

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
HendrikLeuschner: Okay.
Other question: This method checks for `access`, `speed` and traffic. The method before, `hasTimeDependentSpeed`, does not check `access`. As far as I can tell, `hasTimeDependentSpeed` is used to set `TD_ASTAR` as calculation algorithm in `ORSGraphhopper.calcPath`. Why does having time dependent access not lead to use of `TD_ASTAR`?

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/isochrones/ResultTest.java
HendrikLeuschner: This tests the resulting size of the isochrone polygon when smoothing is set to 10. What do you suggest it should test instead? And should we not test smoothing results?

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
HendrikLeuschner: There is no variable here to `assert`. It is a comment explaining why a parameter in the following constructor is set the way it is. Not new code either, but that doesn't really matter.

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
HendrikLeuschner: I agree it is not great in here. Making this configurable via app.config is not a useful solution either, as this specifically only applies to traffic given within _one_ time zone, which means if we want to make it configurable for other traffic data sets, we need to take that circumstance into account, too. Any good idea how to configure this?

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
HendrikLeuschner: Done

openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
HendrikLeuschner: Refactoring old code was not part of this development. The code has been like this for years. Needs proper refactoring at some point, which would be more in-depth and universal.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/TrafficGraphStorage.java
HendrikLeuschner: @MichaelsJP

openrouteservice/src/main/java/org/os_concavehull/ConcaveHull.java
HendrikLeuschner: Removed the whole package as it is not used. It supported holes for isochrones, but this was not tested and adjusted thoroughly enough imo. Also unnecessary at the moment. Can be re-added if ever wanted.

openrouteservice/src/main/java/org/os_concavehull/ConcaveHull.java
HendrikLeuschner: Removed

openrouteservice/src/main/java/org/os_concavehull/TriCheckerAlpha.java
HendrikLeuschner: There is no library for this. But code is removed now anyway.

openrouteservice/src/test/java/org/heigit/ors/routing/traffic/TrafficSpeedCalculatorTest.java
HendrikLeuschner: Don't think this changed anything about readability or performance, but I replaced it.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/TrafficGraphStorage.java
MichaelsJP: @HendrikLeuschner  @sfendrich I moved TrafficRelevantWayTypes and RoadTypes to enums  and refactored their usages

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
sfendrich: O.K., I see. Maybe you could at least add an explanation why this is so?

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/isochrones/ResultTest.java
sfendrich: The question is, where do the numbers 51 and 49 come from? A test should capture a requirement. If the test-result varies with the implementation details, then the test does not check for a correct result but for a random result of the variation. Is there any stable property we would expect, when smoothing is set to 10?

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
sfendrich: Maybe as a property of the storage or the the used encoder?

openrouteservice/src/test/java/org/heigit/ors/routing/traffic/TrafficSpeedCalculatorTest.java
sfendrich: Thanks. Concerning readability: With a switch you immediately know that the variable you are branching on is always the same and you communicate that this is the intended bahaviour. In an if-else-chain you have to check each if-statement to know this fact. Concering performance: the compiler creates a lookup-table to know where to jump to when switch is used. In the below example if edgeId=6 the processor must perform all previous if-instructions.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/TrafficRelevantWayType.java
MichaelsJP: Are gone now.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/TrafficRelevantWayType.java
MichaelsJP: gone

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
HendrikLeuschner: Not really, because this is not my code. It was written by Max I guess. And just from looking at it, I do not know why it only supports node based. Extends Dijkstra, and the extensions do not seem like they necessarily need to be node based to me. But I know as much as anyone else about that.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/reader/traffic/HereTrafficReader.java
MichaelsJP: @sfendrich please give me a bit more details on your thoughts on this. thx.

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/isochrones/ResultTest.java
HendrikLeuschner: We can test the condition that stronger smoothing results in fewer coordinates for the polygon. That's the stable property of smoothing if you want to be implementation independent.

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
HendrikLeuschner: done

openrouteservice/src/main/java/org/heigit/ors/isochrones/GraphEdgeMapFinder.java
sfendrich: Ok, then I suggest to would add a comment
`// TODO: understand reason for TraversalMode restriction`

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/reader/traffic/HereTrafficReader.java
sfendrich: It is difficult to read and reason about deeply nested control flow structure (for, if, ...). Therefore, I would split up this method into several smaller private methods with well-chosen names, which explain what happens. Imagine the top-level method as a table of contents only mentioning the major steps. Each step is detailed out in a private method like a subsection in a book.

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/isochrones/ResultTest.java
sfendrich: Yes, that might be a better test

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
aoles: Good point! In the current implementation selecting `TD_ASTAR` would trigger the use of `TDCoreALT`. The problem with this approach is that at the moment time-dependent access restrictions are not entirely compatible with computing the landmarks in core. As far as I remember during landmark computation the default state of edges with time-dependent access is not blocked but rather uses the default access value same as for non-time dependent routing. These might violate the approximator's correctness in the case when some closed edges become open during routing. So long story short, time-dependent access restrictions are not supposed to be used with `TDCoreALT` yet. AFAIU this shouldn't be the problem for our customer as they don't use OSM time dependent speed or access restrictions, just the variable speed based on third-party traffic data.

aoles: I've simplified and renamed these methods in order to clarify the actual conditions.

openrouteservice-api-tests/src/test/java/org/heigit/ors/v2/services/isochrones/ResultTest.java
HendrikLeuschner: done

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/reader/traffic/TrafficLinkMetadata.java
HendrikLeuschner: @MichaelsJP

HendrikLeuschner (approved):

openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
HendrikLeuschner: Refactoring old code was not part of this development. The code has been like this for years. Needs proper refactoring at some point, which would be more in-depth and universal.

openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
HendrikLeuschner: Refactoring old code was not part of this development. The code has been like this for years. Needs proper refactoring at some point, which would be more in-depth and universal.

openrouteservice/src/main/java/org/heigit/ors/mapmatching/hmm/HiddenMarkovMapMatcher.java
HendrikLeuschner: @MichaelsJP ? Guess this is old code, too.

openrouteservice/src/main/java/org/heigit/ors/mapmatching/hmm/HiddenMarkovMapMatcher.java
HendrikLeuschner: Same as above

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
HendrikLeuschner: Not a part of this PR. If you want to change existing code, this is not the place for it. I'd suggest this has to be done in THE refactoring that is supposed to happen at some point.

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
HendrikLeuschner: Same as above

openrouteservice/src/main/java/org/heigit/ors/routing/RoutingProfile.java
HendrikLeuschner: Same as below. Changing this here makes no sense. Either change it everywhere in a proper comment-formatting PR, or as part of some refactoring process. Changing it here now is inconsistent and has nothing to do with the scope of this PR.

openrouteservice/src/main/java/org/heigit/ors/isochrones/builders/concaveballs/ConcaveBallsIsochroneMapBuilder.java
sfendrich: O.K. let's do this separately, although I am not a fan of big refactoring sessions. Continuously moving towards better code is usually much easier.

MichaelsJP:
:champagne:  :fireworks: :man_dancing:

PR #1057, opened 2021-09-14
Changing BUILD_GRAPHS to ENV

jimbo8098:
Based on comments in #967 , looking to change BUILD_GRAPHS to an environment variable instead of a build argument. This way someone can set the environment variable at container start and it will clear the graph dir. With `ARG` this would, theoretically, only remove the graphs when they don't exist if this build directory is empty at build time.

Kudos to @joker234 for the suggestion.


takb (approved):
lgtm

PR #1130, opened 2022-02-22
Adapt LocationIndex query in centrality to GH4.0

koebi:
The LocationIndex.query was changed from using a node-based to using an edge-based approach.
Nodes therefore have to be inferred from the edges.


takb (approved):
lgtm

PR #1136, opened 2022-03-02
Rphast gh update

HendrikLeuschner:
Fix RPHAST to be runnable and callable. Some API tests still failing, so need to check those still.


sfendrich (approved):
LGTM

PR #1147, opened 2022-03-29
Add Comparison test for CH and RPHAST, but leave commented out

HendrikLeuschner:
Please check whether the changes to pom are ok before merge. Removed junit and added jqwik instead.
Also changed maven surefire version.

Leaving the test commented out because there are some runs when there is a failure, which still needs to be investigated.


jlink (approved):
The changes in JUnit, JUnit Platform and jqwik dependencies look good to me.

takb (approved):
LGTM

Issue #1165, opened 2022-05-11
No differentiation between de- and incline in wheelchair routing


koebi:
#### [Old] Parse incline tag "down" as decline 

This happens for the wheelchair profile at 

https://github.com/GIScience/openrouteservice/blob/76d7c97761bdf768d22db5157332d8b9e25823c5/openrouteservice/src/main/java/org/heigit/ors/util/UnitsConverter.java#L188

where an incline tag of "down" is replaced with "10".

I'd expect it to be replaced by "-10", as it designates a decline?

koebi:
In general, the description and fix seem to be the correct.

Nevertheless, this will be put on hold for the following reason:

The containing method is only called once in the whole codebase. There, `declineToIncline` is set to true, so this change wouldn't affect anything (since the returned value is 10 in either case). Currently, the wheelchair profile only has a `maximum_incline`-method. For this change to be effectively usable, the wheelchair profile would have to differentiate of decline and incline, and this would have to be thoroughly tested. 

### Regarding the new issue:

There is neither a test for the containing method, nor does a test for wheelchair routing exist that even remotely covers complete behaviour of the logic surrounding the `maximum_incline` parameter. This would have to be written from scratch.

It's unclear whether differentiating decline and incline is even desirable, since it is very well possible that it wouldn't make a difference for wheelchair users.

The correct way forward here is to fix the code (which has been happened) and to then reflect the differentiation between de- and incline in the wheelchair profile.

PR #1168, opened 2022-05-12
Parse incline tag "down" as decline

koebi:
This is related to #1165.
As stated there, this change seems to be correct, yet it will not have any effect on the existing behavior.
For completeness, I'd still like to see this merged.

Once that is done, I will change the corresponding issue to track the underlying idea of differentiating between `maximum_incline` and `maximum_decline` in the wheelchair profile and put it on hold.

Is an entry to the changelog neccessary, and if so, what should it look like?


takb (approved):
lgtm, thx

PR #1203, opened 2022-08-02
added Romanian and Esperanto Language

ecxod:
corrected some JSON syntax  
added Romanian and Esperanto Language


takb (approved):
Thanks for your contribution! Since adding languages also requires changes to some API enums as well, I will merge this to a separate branch and merge to master in an additional step.

PR #1234, opened 2022-11-30
make docker build multistage

### LINKED ISSUE #1232: multi-stage ORS image
Partly my own fault I guess, but currently the ORS docker image takes _ages_ to compile ORS every time a container is started. Very annoying when it's supposed to run in CI to validate libraries against. 

The image should compile the code, not the container, so I quickly multi-staged it to only carry over the WAR to the final image. The image size is around half of what it was before which still seems way too big, will see why that is tmrw. Container startup is at least immediate. Will provide a PR soon.

nilsnolde:
Fixes #1232 

Multistaging the ORS image brings down the uncompressed size from 1.06 GB to 365 GB.

EDIT: actually the most important improvement I forgot😅. Before, the Java code was compiled for _every_ container. Now it’s compiled in the image, and a container starts building a graph instantly, instead of first compiling the project for 10 mins.

I also tried for along time to keep the logic with `USER` etc, but eventually it seems rather like it was more redundant and didn't really do what I'd have expected it to do, even before the change. So I removed most of that which makes it a bit easier to maintain I guess.

Basically, before this change, the paths which are mounted as volumes in the docker-compose, were never really owned by whatever the UID was. They were always owned by `root`. If I actually provided a user ID in docker-compose.yml that matched my host one (1000), it wasn't able to create any files in those directories. AFAIU it's happening because after the image is created, those paths are mounted on the host machine, which makes them `root`. I tried fixing that as well, but couldn't in any reasonable time. TBH, I do that exact thing on our [Valhalla image](https://github.com/gis-ops/docker-valhalla) and it works. I tried replicating exactly that for 2 hours with this image and I couldn't get it to work. In the end the effect is that nothing changes compared to before, other than having 1/3 of the image size and a MUCH quicker container startup time.

More comments in the changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:11-jdk
+FROM maven:3.8-jdk-11-slim as build
 
 ENV MAVEN_OPTS="-Dmaven.repo.local=.m2/repository -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=WARN -Dorg.slf4j.simpleLogger.showDateTime=true -Djava.awt.headless=true"
 ENV MAVEN_CLI_OPTS="--batch-mode --errors --fail-at-end --show-version -DinstallAtEnd=true -DdeployAtEnd=true"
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,60 +1,54 @@
-FROM openjdk:11-jdk
+FROM maven:3.8-jdk-11-slim as build
```
nilsnolde, line 1: so we don't have to install maven

Dockerfile
```diff
@@ -1,60 +1,54 @@
...
 
 # Install dependencies and locales
 RUN apt-get update -qq && \
-    apt-get install -qq -y locales nano maven moreutils jq && \
-    rm -rf /var/lib/apt/lists/* && \
-    locale-gen en_US.UTF-8
-
-USER ors:ors
-WORKDIR /ors-core
+    apt-get install -qq -y nano moreutils jq wget
```
nilsnolde, line 15: inside the final image we have now these:
```
> locale -a
C
C.UTF-8
POSIX
```

Dockerfile
```diff
@@ -1,60 +1,54 @@
...
 
-COPY --chown=ors:ors openrouteservice /ors-core/openrouteservice
-COPY --chown=ors:ors $OSM_FILE /ors-core/data/osm_file.pbf
-COPY --chown=ors:ors $ORS_CONFIG /ors-core/openrouteservice/src/main/resources/ors-config-sample.json
-COPY --chown=ors:ors ./docker-entrypoint.sh /ors-core/docker-entrypoint.sh
+COPY openrouteservice /ors-core/openrouteservice
+COPY $OSM_FILE /ors-core/data/osm_file.pbf
+COPY $ORS_CONFIG /ors-core/openrouteservice/src/main/resources/ors-config-sample.json
```
nilsnolde, line 19: no USER necessary in the build image

Dockerfile
```diff
@@ -1,60 +1,54 @@
...
     cp -R /tmp/apache-tomcat-${TOMCAT_VERSION}/* /usr/local/tomcat/ && \
     rm -r /tmp/tomcat.tar.gz /tmp/apache-tomcat-${TOMCAT_VERSION}
 
-# Configure ors config
+# Configure ors config:
+# - Replace paths in ors-config.json to match docker setup
+# - init_threads = 1, > 1 been reported some issues
+# - Delete all profiles but car
 RUN cp /ors-core/openrouteservice/src/main/resources/ors-config-sample.json /ors-core/openrouteservice/src/main/resources/ors-config.json && \
-    # Replace paths in ors-config.json to match docker setup
```
nilsnolde, line 42: docker interprets them as empty lines and raises a warning in newer versions.

Dockerfile
```diff
@@ -1,60 +1,54 @@
...
     jq '.ors.services.routing.profiles.default_params.graphs_root_path = "data/graphs"' /ors-core/openrouteservice/src/main/resources/ors-config.json |sponge /ors-core/openrouteservice/src/main/resources/ors-config.json && \
-    # init_threads = 1, > 1 been reported some issues
     jq '.ors.services.routing.init_threads = 1' /ors-core/openrouteservice/src/main/resources/ors-config.json |sponge /ors-core/openrouteservice/src/main/resources/ors-config.json && \
-
-    # Delete all profiles but car
     jq 'del(.ors.services.routing.profiles.active[1,2,3,4,5,6,7,8])' /ors-core/openrouteservice/src/main/resources/ors-config.json |sponge /ors-core/openrouteservice/src/main/resources/ors-config.json
 
-# Make all directories writable, to allow the usage of other uids via "docker run -u"
-RUN chmod -R go+rwX /ors-core /ors-conf /usr/local/tomcat /var/log/ors
+RUN mvn -f /ors-core/openrouteservice/pom.xml package -DskipTests
```
nilsnolde, line 40: build the project here, then just copy the artifact over to the runner

Dockerfile
```diff
@@ -1,60 +1,54 @@
...
-VOLUME ["/ors-core/data/graphs", "/ors-core/data/elevation_cache", "/ors-conf", "/usr/local/tomcat/logs", "/var/log/ors"]
+COPY --from=build /ors-core/openrouteservice/target/ors.war .
+COPY --from=build /ors-core/openrouteservice/src/main/resources/ors-config.json .
+COPY --from=build /usr/local/tomcat /usr/local/tomcat
+COPY ./docker-entrypoint.sh .
 
 # Start the container
 EXPOSE 8080
-ENTRYPOINT ["/bin/bash", "/ors-core/docker-entrypoint.sh"]
+CMD ["/bin/bash", "/ors-core/docker-entrypoint.sh"]
```
nilsnolde, line 54: IMO `CMD` makes more sense here

docker-entrypoint.sh
```diff
@@ -22,21 +25,18 @@ echo "### openrouteservice configuration ###"
...
-	cp -f /ors-core/openrouteservice/target/*.war /usr/local/tomcat/webapps/ors.war
+	cp -f /ors-core/ors.war /usr/local/tomcat/webapps/ors.war
 fi
 
-/usr/local/tomcat/bin/catalina.sh run
-
-# Keep docker running easy
-exec "$@"
+# so docker can stop the process gracefully
+exec /usr/local/tomcat/bin/catalina.sh run
```
nilsnolde, line 42: Without exec'ing here, docker takes 10 secs to stop a container, since it'd do it in another shell.

rabidllama:
Seems OK to me, the main thing though that I am not sure about is the user stuff as from what I recall, that was added to fix problems people were having with the container not being able to write to the folders

nilsnolde:
The way I read the code before these changes was that you'd provide a host's UID & GID per env vars which are then used for the container-internal `ors` user, and default to `0`, ie root: https://github.com/GIScience/openrouteservice/blob/338979f5e2c7bb26fa157f6019d54e09c68db637/docker/docker-compose.yml#L14

The idea behind that is usually that you'd like to avoid having to use `sudo` on the host for whatever the container creates and is mapped back to the host (plus the docker internal user is not `root` which is also nice when exposing as host volumes). However, I don't think that was ever properly working and my attempt to make it work wasn't very successful.

Anyways, you can take whatever you think is nice from this PR and trash the rest. I'm only using this image to test some of our routing py/js libs and I'm happy with our custom image, likely for years to come.

Dockerfile
MichaelsJP: Nice. Well done. 
While you're at it :), please pin the package versions from all packages installed in the image and add `--no-install-recommends`, as described here: 

https://github-wiki-see.page/m/hadolint/hadolint/wiki/DL3008
https://github-wiki-see.page/m/hadolint/hadolint/wiki/DL3015

In my opinion, the versions don't really need their own ARG, but can be hard-coded.

Dockerfile
MichaelsJP: Nice catch. Besides, it could be worth adding `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` before the run, since we have plenty of pipes in it. Else, errors during the piping are never caught, except the last exit code. For reference:

https://github-wiki-see.page/m/hadolint/hadolint/wiki/DL4006

MichaelsJP:
@nilsnolde Thanks for your contribution. More than welcome. I left you some comments. Just leave me a little comment if you want to provide the changes, else we'd happily just take your commits and do the rest ourselves. :)

MichaelsJP:
Cherrypicked your commits into a new branch and made a new PR: https://github.com/GIScience/openrouteservice/pull/1272

Will be merged soon.

PR #1284, opened 2023-02-16
Fix docker build issues

MichaelsJP:
OpenSSL has some vulnerabilities that are fixed in 1.1.1t-r0 →  https://www.openssl.org/news/openssl-1.1.1-notes.html

In the same course, the glibc is also raised to its latest version.


takb (approved):
lgtm

PR #1290, opened 2023-02-22
Isolate workflows for nightly and version docker images

MichaelsJP:
The current workflows do not cover scenarios where a release triggers the release workflow instead of a merge request with an included release tag.


takb (approved):
Sounds good, should makes it clearer

PR #1328, opened 2023-03-14
doc(GTFS): profile name in api docs CUSTOM_KEYS

takb:
- Reason for change:
Old profile names were still used in the CUSTOM_KEYS references

### DIFF
--- a/openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequest.java
+++ b/openrouteservice/src/main/java/org/heigit/ors/api/requests/routing/RouteRequest.java
@@ -281,7 +281,7 @@ public class RouteRequest extends APIRequest {
      * Other parameters public-transport accepts are coordinates and language.
      */
     @ApiModelProperty(name = PARAM_SCHEDULE, value = "If true, return a public transport schedule starting at <departure> for the next <schedule_duration> minutes." +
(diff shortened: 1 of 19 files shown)


koebi (approved):

TheGreatRefrigerator:
i still found these, do they also need adjustment?:
- [ ] https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/resources/resources/locales/ors_de-DE.resources#L94
- [ ] https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/resources/resources/locales/ors_en-US.resources#L94
- [ ] https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/org/heigit/ors/api/responses/routing/json/JSONLeg.java#L38
- [ ] https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/org/heigit/ors/api/responses/routing/json/JSONLeg.java#L112


For the locales, only US and DE are added for the public transit, not sure if there are other languages should also have the `pt` object.

takb:
I don't think we need to use the full profile name for these four strings you mention, the language file key is something internal and the leg type value for the responses only produce unnecessary bloat. 

> For the locales, only US and DE are added for the public transit, not sure if there are other languages should also have the `pt` object.

But you are right about this, I'll add untranslated values to the other language files. Thx for the heads-up!

PR #1356, opened 2023-03-22
Add war deployment check

### LINKED ISSUE #1350: test war-file deployment
the war-file deployment is currently not tested but should be

### LINKED ISSUE #1355: Add war deployment test
Currently, the war deployment is not tested. This could be easily done by running the docker image that is anyway built.

MichaelsJP:
Currently, the war deployment is not checked. This starts the container and waits for the health API to indicate the default graph is built and ready.

### Pull Request Checklist

         have been resolved.

Closes #1355 

- Key functionality added: Test for correct war deployment.

### DIFF
--- a/.github/utils/health_check.sh
+++ b/.github/utils/health_check.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+host=${1}
+port=${2}
+base_path=${3}
+
+wait_for_url() {
+    echo "Waiting for status code 200: $1"
+    # shellcheck disable=SC2016
+    timeout --foreground -s TERM 240s bash -c \
+        'while [[ "$(curl -s -o /dev/null -m 3 -L -w ''%{http_code}'' ${0})" != "200" ]];\
+        do echo "Waiting for ${0}" && sleep 2;\
+        done' "${1}"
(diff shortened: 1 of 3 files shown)


jlink (commented):
Does it work - asking because the build step seems to be hanging.

I don't understand all the details of the new build step, what strikes me is that the test will only be run on CI. That's rather late feedback. If that's ok, depends on how often we expect a problem with the war deployment.

MichaelsJP:
@jlink 

> Does it work - asking because the build step seems to be hanging.

Yes. The action name has changed, and it's still looking for the old. Will change that before the merge. Else, all other PRs will have to wait for it.

> I don't understand all the details of the new build step, what strikes me is that the test will only be run on CI. That's rather late feedback. If that's ok, depends on how often we expect a problem with the war deployment.

I'm open to ideas on how to solve that deployment scenario locally, as isolated as it is in the workflow with the Dockerfile. In my opinion, it's ok to have that kind of feedback latest with a PR, so there is no danger of merging it into master. It also runs on a draft PR, so feedback is given quick if commits are uploaded regularly.

jlink (approved):
Let's start with your approach and think about a way to extract that from GH actions later.

PR #1376, opened 2023-03-30
Feat: reintegrate traffic storage

### LINKED ISSUE #1343: Merge time-dependent and traffic
#1339 
#1341 
#1340 

- [x] Write a minimal API test
- [x] Reintegrate traffic storage
- [x] Create freely publishable test-dataset for Heidelberg
- [ ] Find routes and isochrones that differ due to traffic and create a suitable API test
- [ ] Enable traffic reading in config
- [ ] Enable API tests with more detailed ones for the td
- [ ] Implement @aoles commets ☄️

### LINKED ISSUE #1359: Re-enable traffic integration
- [ ] Move the matching hook to GraphHopper.
- [ ] Sort out why the map matcher gives errors.

sfendrich:
The traffic storage has not been carried over to ORS 7. This commit reintegrates the reading and storing of traffic data. The reading of traffic data is disabled in the config and like-wise the corresponding API test because we currently have no freely publishable test data.

### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Relates to [#1343](https://github.com/GIScience/openrouteservice/issues/1343)

- Key functionality added: Reintegration of traffic storages
- Reason for change: enable traffic dependent routing


openrouteservice/src/test/java/org/heigit/ors/apitests/isochrones/ResultTest.java
```diff
@@ -478,4 +479,23 @@ void testCompleteMetadata() {
                 .body("metadata.containsKey('system_message')", is(true))
                 .statusCode(200);
     }
+
+    @Disabled
+    @Test
+    void testAcceptsTimeParameter() {
```
aoles: AFAIU the purpose of this test is to assess whether providing the time parameter in the case traffic data is enabled yields a valid response, and does not care about the actual content of the response. As such, it might be better harbored under `ParamsTest`, what do you think?

openrouteservice/src/test/resources/ors-config.json
```diff
@@ -192,6 +192,15 @@
                 },
                 "RoadAccessRestrictions": {
                   "use_for_warnings": true
+                },
+                "HereTraffic": {
+                  "enabled": false,
+                  "streets": "src/main/files/here_traffic/germany_traffic/Streets.shp",
```
aoles: I believe files which are not being included in the repo shouldn't be referenced in the config, unless I'm missing sth here?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
```diff
@@ -401,11 +404,29 @@ private LineString constructFreeHandRouteGeometry(GHRequest request) {
...
+            }
+        }
+    }
+
     /**
 	 * Does the preparation and creates the location index
 	 */
 	@Override
 	protected void postProcessingHook(boolean closeEarly) {
+		matchTraffic(); // TODO: Must this be moved into GraphHopper like it was originally?
```
aoles: Re the TODO:  depending on whether traffic data is to be used with TD ALT, it might actually be necessary to have the data available before creating the landmarks. This would be in particular true when traffic speed could override the default edge-speed by setting it to a higher value. In such scenario the maximum speed value needs to be considered for calculating LMs in order to ensure the correctness of the approximator.

aoles (changes requested):
Thanks, please consider my inline comments. Cheers!

openrouteservice/src/test/java/org/heigit/ors/apitests/isochrones/ResultTest.java
sfendrich: I don't know. Looking through existing tests, the distinction between Params and Results seems arbitrary. I don't mind whether this test is here or there. I can put it where you wish.

openrouteservice/src/test/resources/ors-config.json
sfendrich: This is one of the missing mock files that should be added.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
sfendrich: So, is there anything to do for now?

openrouteservice/src/test/java/org/heigit/ors/apitests/isochrones/ResultTest.java
aoles: Thanks, maybe @jlink has some thoughts on this?

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
aoles: According to the implementation of `TrafficSpeedCalculator` traffic speed can actually override the edge speed by setting it to a higher value in [here](https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/org/heigit/ors/routing/traffic/TrafficSpeedCalculator.java#L53-L66). Therefore, it is important that the landmarks are build only after traffic data has been loaded such that they can use use the maximum possible speed taking into account free flow traffic as in [here](https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/org/heigit/ors/routing/traffic/TrafficSpeedCalculator.java#L44-L47).

openrouteservice/src/test/java/org/heigit/ors/apitests/isochrones/ResultTest.java
jlink: Maybe there exists a better criterion than Param/Result for chunking the tests. What I do know is that many of the ResultTest classes have way too many tests to find my way around.

aoles: Not sure what the best strategy is. As long as the test names are clear and descriptive enough I don't mind having the tests aggregated in classes grouped by endpoints. For now I would suggest we try to stick to what is already there and try to be consequent when introducing new stuff.  FWIU the distinction between `Param/Result` tests is that the former ones validate whether certain parameters yield meaningful response while the latter one actually deals with the correctness of the results. But of course, there is no clear distinction, and each one of use might have a slightly different perception where the borderline is. From my point of view it makes sense to move the test unless someone objects.

sfendrich: O.K. I have moved the test into ParamsTests and created an issue on the cleanup project board.

openrouteservice/src/test/java/org/heigit/ors/apitests/isochrones/ParamsTest.java
```diff
@@ -18,6 +18,7 @@
 import org.heigit.ors.apitests.common.VersionAnnotation;
 import org.json.JSONArray;
 import org.json.JSONObject;
+import org.junit.jupiter.api.Disabled;
```
aoles: Seems to be unused.

openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
sfendrich: Moved matchTraffic into graphhopper

openrouteservice/src/test/java/org/heigit/ors/apitests/isochrones/ParamsTest.java
sfendrich: Oh, sure, the test is enabled now. Removed.

aoles (approved):

aoles:
Thanks @sfendrich 🎉

PR #1380, opened 2023-03-31
feat(swagger): Springfox to openapi migration

### LINKED ISSUE #1027: migrate from springfox-swagger to springdoc-openapi
Currenlty swagger 2 is still used as API specs.
OpenAPI specs have been out for some time an are the go to standard nowadays.

It might be possible to integrate the custom keys used to define parameter dependencies in a more openapi spec native way (to be investigated).

Should be tackled together with [ors-website's capability to render openapi](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/openrouteservice-infrastructure/ors-website/-/issues/106).

MichaelsJP:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Close #1027 

- Key functionality added: Upgrade from springfox to springboot openapi 
- Reason for change: The springfox package is not maintained and outdated. It's also not compatible with spring-boot 3, which is the only blocker currently to not upgrade.

- Added `swagger_documentation_add_autogenerated_url` to let the user decide if the auto-generated (localhost) url should be added to the available servers. This is needed since the openapi package defines the servers as an array of possible endpoints, rather than a fixed string for one url only.


openrouteservice/src/main/java/org/heigit/ors/api/OpenApiConfiguration.java
```diff
@@ -0,0 +1,104 @@
...
+            Server localhostServer = new Server().url(servletContext.getContextPath()).description("Auto generated server url");
+            listOfServers.add(localhostServer);
+        }
+        return listOfServers;
+    }
+
+    private Info apiInfo() {
+        return new Info()
+                .title(SERVICE_NAME)
+                .description("This is the openrouteservice API documentation")
```
TheGreatRefrigerator: If it's possible somehow it would be great to have the ors version tag dynamically provided here.
We currently do this in the api-doc service:
https://gitlab.gistools.geog.uni-heidelberg.de/giscience/openrouteservice-infrastructure/ors-docs-api/-/blob/main/cron/docs_update_check.py#L48-65

MichaelsJP: > ors_live_version = version.parse(status.get('engine').get('version'))

You're referring to the ors engine version?
As I understand the version tag in the `apiInfo` object, it's targeted towards the API version (v1/v2/v3).

As stated in the [swagger docs](https://swagger.io/docs/specification/api-general-info/) we could use a versioning approach. E.g. `version: v2-7.0.1` or vice versa.

That way we could always extract the ors version from the docs and also know with what ors version they were generated.

MichaelsJP: We could dynamically adjust the description part with the normal version.
An edge case would be if ors is run from master (nightly). -> This also applies to /status in effect, so we could also forget about it ;).

openrouteservice/src/main/java/org/heigit/ors/api/OpenApiConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+            listOfServers.add(localhostServer);
+        }
+        return listOfServers;
+    }
+
+    private Info apiInfo() {
+        return new Info()
+                .title(SERVICE_NAME)
+                .description("This is the openrouteservice API V2 documentation for ORS Core-Version " + VERSION  + ". Documentations for [older Core-Versions](https://github.com/GIScience/openrouteservice-docs/releases) can be rendered with the [Swagger-Editor](https://editor-next.swagger.io/).")
+                .version("2")
```
MichaelsJP:
```suggestion
                .version("v2")
```

jarinox:
There is a `ptRequest` field generated for `directions/*` POST routes which should not be there.  

image

MichaelsJP:
@jarinox Yes and no. We'll have public transit soon in our live API and the code supports it already. A curl request looks like this

```
curl -d '{"coordinates": [[8.6729581, 49.4468535], [8.7067204, 49.3786147]], "departure": "2022-07-04T13:02:26Z"}'` \
-H "Content-Type: application/json" -X POST http://localhost:8080/ors/v2/directions/pt
````
Makes sense with the parameter, right?
But maybe it could be worth hiding it from our live API docs until it's really available.

takb (approved):
finally! lgtm.

PR #1397, opened 2023-04-12
fix (unit tests): Make CountryBordersReaderTest reset state

jlink:
## Fixes failure of routing.ResultTest.testCountryExclusion when run in full test suite - and only then.

The root cause is that CountryBordersReader internally uses a singleton.
Never do that! Never do that! Never do that! Never do that! Never do that!
Singletons are bad! Singletons are bad! Singletons are bad! Singletons are bad! Singletons are bad!

The fix does not get rid of the singleton, which would be quite a refactoring. It just resets it to its previous state in the unit test that changes it.


aoles (approved):
Awesome, thanks a lot @jlink! I can confirm that this fixes the original issue.

PR #1402, opened 2023-04-18
Issue#1391:  support docker images for arm64

### LINKED ISSUE #1391: High graph building/response time on Apple Silicon Processors
While trying to start custom ORS docker image (built on top of the public image), on Silicon M1 Macbook Pro, we face too long graph building times for a 2GB osm.pbf file.

The same process on an EC2 ubuntu instance, with even lower specs, takes about 10 minutes, using the same Xms and Xmx heap size (`-Xms5g -Xmx7g` for 2GB osm.pbf file)

And even when the graphs are finally built and the service is up and running, the reponse times are higher than on the ubuntu EC2.
Given this difference in performance, we suspect that the delay is caused by docker engine running in compatibility/emulation mode on the macbook pro, as the ORS images are built only for `linux/amd64` architecture.

---

A solution to this, is updating ORS' CI pipeline to build and publish images for `linux/arm64` too,  apart from `linux/amd64` architecture only.
This will remove the need to run the images on Silicon processors in compatibility/emulation mode.

---

It is easy to change this by updating the following two variables on CI:

* [BUILD_PLATFORMS](https://github.com/GIScience/openrouteservice/blob/master/.github/workflows/build-and-publish.yml#L12)

* [DOCKER_PLATFORMS](https://github.com/GIScience/openrouteservice/blob/master/.github/workflows/publish-tagged-release.yml#L47)

from `linux/amd64` to `linux/amd64,linux/arm64`

The final image will be compatible with both platforms, with the correct version automatically pulled

rolkool:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes #1391 .

Adds support for arm64 docker images (e.g: for Apple Silicon M1 processors)

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -54,7 +54,7 @@ jobs:
           load: true
           tags: ${{ needs.prepare_environment.outputs.test_image_name }}
           #build-args: "--platform ${{ needs.prepare_environment.outputs.build_platforms }}"
-          platforms: "${{ needs.prepare_environment.outputs.build_platforms }}"
+          platforms: 'linux/amd64'
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Start container from image and wait for successful checks
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -35,7 +35,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+        uses: docker/setup-qemu-action@v2
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
         id: buildx
@@ -87,7 +87,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+        uses: docker/setup-qemu-action@v2
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
         id: buildx
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -53,7 +53,8 @@ jobs:
           push: false
           load: true
           tags: ${{ needs.prepare_environment.outputs.test_image_name }}
-          build-args: "--platform ${{ needs.prepare_environment.outputs.build_platforms }}"
+          #build-args: "--platform ${{ needs.prepare_environment.outputs.build_platforms }}"
+          platforms: "${{ needs.prepare_environment.outputs.build_platforms }}"
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Start container from image and wait for successful checks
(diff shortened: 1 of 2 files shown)


MichaelsJP:
@rolkool Thanks for opening the PR :). We'll review it later today.

MichaelsJP:
The problem is probably that the image is built with `arm64` and `--load` on a non arm64 platform. `--load` is needed to be able to run the docker image in the workflow.

A solution could be to have to buildx calls. 

I'll need to spend a few thoughts on this.

MichaelsJP:
Ok. Just had an idea. Best would be to change the build process into a matrix build strategy. What I mean can be seen here for example: https://github.com/Overv/openstreetmap-tile-server/actions/runs/2694331977/workflow.

We could adapt that.

To be able to also start the runner with the right platform we need this crucial task
```yaml
-     name: Set up QEMU
      uses: docker/setup-qemu-action@v2
      with:
        platforms: ${{ matrix.arch }}
```

rolkool:
@MichaelsJP I think we can use only amd64 for build step, and both platforms for `publish_docker` job

MichaelsJP:
> @MichaelsJP I think we can use only amd64 for build step, and both platforms for `publish_docker` job

Generally yes, but I still think it's important to see the docker checks run on all platforms that we build images for. Not only amd64.

MichaelsJP:
I mean this part specifically, that should be running and passing on all platforms
```yaml
      - name: Start container from image and wait for successful checks
        run: |
          mkdir $(pwd)/graphs $(pwd)/conf
          sudo chown 1000:1000 $(pwd)/graphs $(pwd)/conf
          docker run -it -d -p 8080:8080 -v $(pwd)/graphs:/home/ors/ors-core/data/graphs -v $(pwd)/conf:/home/ors/ors-conf --name ors-instance ${{ needs.prepare_environment.outputs.test_image_name }}
          # Check for health to turn 200 after the graphs are build and spring-boot completely started         
          ./.github/utils/url_check.sh 127.0.0.1 8080 /ors/v2/health 200 240
          # Check for correct preflight settings to avoid CORS issues with ORIGIN wildcard from the example config
          ./.github/utils/cors_check.sh 127.0.0.1 8080 /ors/v2/isochrones/geojson "https://example.org" 200 10
          ./.github/utils/cors_check.sh 127.0.0.1 8080 /ors/v2/isochrones/geojson "https://example.com" 200 10
          # Adjust the config file and set ORIGIN to https://example.org
          sudo sh -c 'jq ".ors.api_settings.cors.allowed.origins[0] = \"https://example.org\"" ./conf/ors-config.json | sponge ./conf/ors-config.json'
          # Restart the container to apply the config changes
          docker restart ors-instance
          # Request preflight with https://example.com and https://example.org to see if it gets applied correctly
          ./.github/utils/cors_check.sh 127.0.0.1 8080 /ors/v2/isochrones/geojson "https://example.org" 200 50
          # It should fail with http code 403 for https://example.com since the Origin is not covered.
          ./.github/utils/cors_check.sh 127.0.0.1 8080 /ors/v2/isochrones/geojson "https://example.com" 403 10
```

MichaelsJP:
@rolkool Look here https://github.com/GIScience/openrouteservice/pull/1407
It's implemented now.

MichaelsJP:
@rolkool You should be able to use nightly soon with the latest arm64 platform.

hbilgrami1:
@MichaelsJP I tried to pull the nightly version ` docker pull openrouteservice/openrouteservice:nightly --platform linux/arm64` but there was no matching manifest in the entries. Any thoughts?
`no matching manifest for linux/arm64 in the manifest list entries`

PR #1412, opened 2023-04-25
Fix for `access: destination`

### LINKED ISSUE #253: Rework the motor_vehicle=destination tag penalty logic
Consider driving on the following [field road](https://www.openstreetmap.org/way/157774067).

https://maps.openrouteservice.org/directions?n1=49.490536&n2=8.65129&n3=17&a=49.490507,8.654737,49.490417,8.647077&b=0&c=0&k1=en-US&k2=km

image

The reported driving time for 553 meters is 6 minutes, which translates into a driving speed of about 5 km/h. When switched to `profile=driving-hgv` the time reduces to a more reasonable value of 1 minute. Switching between CH and ALT by enabling some filters does not change these values.

Now, when the distance between the start and end points is slightly increased e.g. by dragging the former point a bit to the left the returned route for `driving-car` switches to the following one; the routing for `driving-hgv` is not affected by this.

https://maps.openrouteservice.org/directions?n1=49.489629&n2=8.650542&n3=16&a=49.490507,8.654737,49.490403,8.646863&b=0&c=0&k1=en-US&k2=km

image

### LINKED ISSUE #705: Use priority weighting to discourage routing over `access=destination` ways
Routing over ways tagged with `access=destination` is currently discouraged by artificially limiting the travel speed to `destination_speed`. The problem with this approach is:
1. It introduces a spurious time penalty for `fastest` weighting reflected in inaccurate travel time estimation
2. It doesn't solve the issue for `shortest` weighting + has the same problem with travel time estimation as for `fastest`.

Investigate whether an approach of penalizing `access:destination` with priority weighting wouldn't be better. The priority of such ways could be set to `WORST`.

Should help to address https://github.com/GIScience/openrouteservice/issues/253.

### LINKED ISSUE #1061: `access=destination` speed of 5 km/h too low?
A user commentary in our [forum](https://ask.openrouteservice.org/t/route-returns-very-slow-driving-rate/3151) reported unusually long drive times on a highway that turned out to be tagged with `access:destination`.

The ors uses a speed of 5 km/h for such routes, compare https://github.com/GIScience/openrouteservice/blob/37fca1d1b6bf1203deb69c4a1ec6633248a725f8/openrouteservice/src/main/java/org/heigit/ors/routing/graphhopper/extensions/flagencoders/VehicleFlagEncoder.java#L157. 

This particular highway has a maxspeed of 45 mph, but the more general question is whether a speed of 5 km/h is justified for roads with `access:destination`.

### LINKED ISSUE #1344: access:destination - remove GH penalty
currently a double penalty both in GH and ORS is done.
Move to previous impl. to reduce differences in results.

- [ ] test locally
- [ ] check output visually
- [ ] merge if ok

touches #1061

aoles:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes #1344
fixes #1061 
fixes #705 
fixes #253 

- [x] fix routing 
- [ ] fix isochrones
- [ ] other occurrences of `FastestWeighting`


sfendrich:
I wonder why the design change of the weighting factory is necessary for introducing the access:destination weighting. Shouldn't these be two separate things?

Further, the "cleanup" feels like a step backwards. We moved away from that design in the past to improve maintainability.

aoles:
Thanks @sfendrich ! Of course, strictly speaking the refactoring is not necessary for the introduction of the additional weighting. However, what was necessary was the introduction of yet another hook in the upstream class in order to override the creation of the `"fastest"` weighting. To be honest, at that point with the rest of modifications to the super class it didn't make much sense to me personally to extend the superclass anymore. FWAIU we moved away from the previous design primarily because back then it was a bit unclear what parts of the original ORS weighting factory were different and necessary compared to the GH's one. Because of the insights we have gained by now I don't think this point is valid anymore. I believe the net outcome of this refactoring is in fact reduced code complexity which should also help maintainability, c.f. https://github.com/GIScience/graphhopper/pull/66.

sfendrich:
The main reason, why we have all these hooks is that ORS is split into two code bases. The refactoring was done to ease merging them, which was the plan at least at that time. So the main question is, how do we want to proceed with the splitted code bases?

aoles:
Sure, but even if the code bases are to be merged, these injection points won't go away or do I miss sth?

aoles:
@sfendrich please let me know how you would like to proceed about this. If you wish, I can roll back all the refactoring. However, this will have the consequence that changes to GH's `DefaultWeightingFactory` will be necessary. These in turn will require releasing an updated GH version for this PR to work. Cheers!

sfendrich:
Whether the injection points go away depends on the injection points and on our goals. We might want to keep some for modularity. However, some of them can easily be removed after the merge (see the comments in the source code of the PR you have linked) because the core classes may be moved into the appropriate packages and will then be accessible from within the core. Changes to GH can then be integrated properly. E.g., moving the weightings into the right package would make them available to the weighting factory, so the changes could be applied there directly. Instead we need hooks to break these dependencies because the ORS  weightings are inaccessible external classes.
If we revert the hooks now, we will have to re-iterate on this after the merge in order to do a proper integration. IMO we could avoid this ping pong between two temporary solutions.

I think this should be decided in the team meeting tomorrow.

MichaelsJP:
@aoles Can we change it to "Ready for review" then?

MichaelsJP:
@aoles Will this be closed as well with this logic?
- https://github.com/GIScience/openrouteservice/issues/1061
- https://github.com/GIScience/openrouteservice/issues/705
- https://github.com/GIScience/openrouteservice/issues/253

aoles:
@MichaelsJP Yes, I hope so 🤞

MichaelsJP (approved):

PR #1432, opened 2023-05-05
ci(cve): Scan the war and docker images with grype

### LINKED ISSUE #1420: Integrate new github actions
There are a few tools around that could make sense to be introduced into our workflows:

- [x] conventional commit check
- [x] conventional pr check
- [ ] grype cve scanner for war and docker

MichaelsJP:
This PR introduces grype as another vulnerability check, next to Snyk. It's being implemented as an informational action right now. It's not as convenient as Snyk, but it seems as it's used frequently in the corporate world. Part of the reasons are, it's open source and way more picky in terms of what is being reported and what not.

The check is executed as the following action:

```yml
- name: Run the Anchore Grype scan action to SARIF
  uses: anchore/scan-action@v3
  id: scan
  with:
    fail-build: false
    output-format: sarif
```
There are two checks now with that action that check for the _docker image_ as well as for the _war file_.

The results are first printed to the runner's log and as well uploaded as a SARIF file with `github/codeql-action`. The results of the SARIF file can then be viewed in the code scanning GitHub overview:

https://github.com/GIScience/openrouteservice/security/code-scanning?query=pr%3A1432+tool%3AGrype+is%3Aopen

Additionally, it runs each Sunday at 00:00 am. 

The goal for now is to just inform us of detected vulnerabilities to be able to react in time.

### DIFF
--- a/.github/workflows/vulnerability-scanning.yml
+++ b/.github/workflows/vulnerability-scanning.yml
@@ -1,0 +1,110 @@
+name: "Vulnerability scan"
+
+on:
+  push:
+    branches: [ "master", "main" ]
+  pull_request:
+    branches: [ "master", "main" ]
+  schedule:
+    - cron: '0 0 * * MON'
+
+permissions:
+  contents: read
+
+jobs:
+  Anchore-War-Build-Scan:
+    name: Analyze maven war file with Grype
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


takb (approved):
Sounds good, though we will have to revisit when we merge the modularization branch.

PR #1458, opened 2023-05-30
ci(docker): Fix the docker build github workflow

### LINKED ISSUE #1338: Rework the docker and docker-compose setup
The docker image already brought improvements in terms of size, stability, and startup time.
What couldn't be solved is the complexity of the setup. It has historically grown to what it is now. 
A new, generally overhauled docker setup would be useful in my opinion. 
This could include:
- Automatic pbf download
- Auto-rebuild on .pbf change
- Auto-rebuild or startup block if the graph is started with a different ors version
- reduce the number of volumes to one or two. A general folder would be easier to handle. The new Dockerfile already stores all relevant tomcat, ors logs, etc. in the /home/ors folder.
- With a combined folder, an osm.pbf could be referenced by env instead of a volume mount.
- ...?

MichaelsJP:
Currently, the workflow doesn't seem to correctly build for arm64 platforms.

The current base image doesn't support arm64 so we need to take another one.
`amazoncorretto:17.0.7-alpine3.17` is the only major java build that supports alpine and arm64. To keep vulnerabilities low we should stick to this. Any debian/ubuntu-based images are just full to the top with CVEs.

Partially fixes #1338  .

- Key functionality added:
- Reason for change:

-

-

### DIFF
--- a/.github/workflows/vulnerability-scanning.yml
+++ b/.github/workflows/vulnerability-scanning.yml
@@ -5,7 +5,10 @@ on:
     branches: [ "master", "main" ]
   pull_request:
     branches: [ "master", "main" ]
-    types: [ opened, synchronize, ready_for_review ]
+    paths:
+      - pom.xml
+      - '**/pom.xml'
+      - Dockerfile
   schedule:
     - cron: '0 0 * * MON'
 
--- a/.github/workflows/vulnerability-scanning.yml
+++ b/.github/workflows/vulnerability-scanning.yml
@@ -5,15 +5,27 @@ on:
     branches: [ "master", "main" ]
   pull_request:
     branches: [ "master", "main" ]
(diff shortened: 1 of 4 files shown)


.github/workflows/vulnerability-scanning.yml
```diff
@@ -5,15 +5,27 @@ on:
     branches: [ "master", "main" ]
   pull_request:
     branches: [ "master", "main" ]
+    types: [ opened, synchronize, ready_for_review ]
```
koebi: Do we really need vuln scanning on *every* commit on *every* PR?
Is this sth that might be enough on "open" and "ready"? Or even on "ready" only?
```suggestion
    paths:
      - pom.xml
      - '**/pom.xml'
      - Dockerfile
```

koebi (approved):

PR #1484, opened 2023-06-28
fix: re-introduce edge metrics cache

aoles:
re-introduce edge metrics cache in matrix path extractor

### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes # .

Matrix path extractor uses a hashmap-based cache to keep track of encountered edges. This is a performance improvement in order to avoid re-computing metrics for edges which have been previously visited. The original solution has been removed in cbecd7d1fa8022e61ae3a0495ed2437550cc543e because it had a flaw of storing the cache entries under non-unique keys computed as sum of edge id and adjacent node id. The caching is reintroduced with a fix of using unique edge keys.

### DIFF
--- a/ors-engine/src/main/java/org/heigit/ors/matrix/PathMetricsExtractor.java
+++ b/ors-engine/src/main/java/org/heigit/ors/matrix/PathMetricsExtractor.java
@@ -13,8 +13,8 @@
  */
 package org.heigit.ors.matrix;
 
+import com.graphhopper.coll.GHLongObjectHashMap;
 import com.graphhopper.routing.SPTEntry;
-import com.graphhopper.routing.util.FlagEncoder;
 import com.graphhopper.routing.weighting.Weighting;
 import com.graphhopper.storage.Graph;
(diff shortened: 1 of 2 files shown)


MichaelsJP (changes requested):
Let's add a short test that tests for the correct caching.
@aoles What do you think?

aoles:
Thanks @MichaelsJP! To be honest I'm not sure if there is a quick and easy way of adding such a test. It will blow up the code to some extent by methods to expose private members of the class. How I see it is that designing and implementing the test alone will probably be more work than the work invested into the fix itself. To put things a bit into perspective, please note that neither `PathMetricsExtractor` nor `DijkstraMatrixAlgorithm` are currently covered by any unit tests, which I personally consider more critical than this one quite specific test. To be fair, however, the algorithm is being called by some of the API tests so it is not completely untested. Having said all this, I would be fine skipping the test in favor of getting more pressing things done.

MichaelsJP (approved):

PR #1488, opened 2023-07-07
refactor: configuration improvements phase 1

takb:
- configuration through application.yml, overridden by application-test.yml/application-unittest.yml for tests and ors-config.yml for actual use
- Properties classes to read said configs
- EngineConfig class to transport what is needed into the engine module
- All previous json config parameters are considered and take precedence
- Moved some API parameters to correct places
- Improved logging with more consistent setup 
- Removed several unused classes

This is a WIP merge, code compiles and runs and all tests succeed, but the new configuration method is not completely functional, especially the profiles configuation part.


MichaelsJP:
Imo “refactor” fits it best right now. When the new config split will be available as the main entrance to the user, I'd say we call that switch "feature".

ors-api/src/main/java/org/heigit/ors/api/servlet/listeners/ORSInitContextListener.java
```diff
@@ -32,17 +34,27 @@
...
+        final EngineConfig config = EngineConfig.EngineConfigBuilder.init()
+                .setInitializationThreads(engineProperties.getInitThreads())
+                .setPreparationMode(engineProperties.isPreparationMode())
+                .setElevationPreprocessed(engineProperties.getElevation().isPreprocessed())
+                .setSourceFile(engineProperties.getSourceFile())
+                .buildWithAppConfigOverride();
         Runnable runnable = () -> {
             try {
-                RoutingProfileManager.getInstance();
+                new RoutingProfileManager(config);
```
MichaelsJP, line 53: :heart_eyes:

MichaelsJP (approved):

PR #1503, opened 2023-07-20
refactor: move logic from Request classes to new service layer

jhaeu:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes # .

- Key functionality added: New service classes with injected EndpointProperties. The logic to create internal requests is moved from APIRequest and its subclasses to the new services. 
- Reason for change: EnspointProperties were only injected to the controllers. We could have passed these properties to the old methods in APIRequest subclasses, or extend the APIRequest classes and set the endpointProperties in the controllers or separate the data of the APIRequest classes and the logic to create internal requests based on their data. I chose the last approach to get a better separation of concerns.

-

-


ors-api/src/main/java/org/heigit/ors/api/services/RoutingService.java
```diff
@@ -0,0 +1,414 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Stream;
+
+
+
+@Service
+public class RoutingService extends AbstractApiService {
+
+    private final EndpointsProperties endpointsProperties;
```
aoles: As all `*Service` classes seem to share a common `private final EndpointsProperties endpointsProperties;`field I would suggest moving it along with the corresponding getter to the parent `AbstractApiService` class.

ors-api/src/main/java/org/heigit/ors/api/services/AbstractApiService.java
```diff
@@ -0,0 +1,424 @@
...
+import org.locationtech.jts.geom.Polygon;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
+public abstract class AbstractApiService {
+
+    abstract EndpointsProperties getEndpointsProperties();
+    abstract double getMaximumAvoidPolygonArea();
```
aoles: All `*Service` classes but `RoutingService` return 0 in `getMaximumAvoidPolygon*` methods. Therefore, you might consider defining this default implementation already here in the upstream abstract class and override it only when relevant.

ors-api/src/main/java/org/heigit/ors/api/services/AbstractApiService.java
```diff
@@ -0,0 +1,424 @@
...
+                    return BordersExtractor.Avoid.ALL;
+                case CONTROLLED:
+                    return BordersExtractor.Avoid.CONTROLLED;
+                default:
+                    return BordersExtractor.Avoid.NONE;
+            }
+        }
+        return null;
+    }
+
```
aoles: Although most of the added code is in fact just code moved from the original classes, I've noticed that in this process some double blank lines have been introduced to separate method bodies. Was this change intentional?

ors-api/src/test/java/org/heigit/ors/api/services/APIServiceTest.java
```diff
@@ -19,56 +21,72 @@
 
 import static org.junit.jupiter.api.Assertions.*;
 
-class APIRequestTest {
-    APIRequest request;
+class APIServiceTest {
+    AbstractApiService apiService;
 
+    static class ApiService extends AbstractApiService {
```
aoles: Might become obsolete once the abstract methods in `AbstractApiService` become default implementations as suggested in my other comments.

aoles (changes requested):
Awesome, thanks a lot @jhaeu for your contribution!

This looks quite good already. Just a few small things to consider, see my inline comments.

On a general note, please go through all of the modified classes and make sure to remove any unused imports 😇

ors-api/src/main/java/org/heigit/ors/api/services/RoutingService.java
jhaeu: will change

ors-api/src/main/java/org/heigit/ors/api/services/AbstractApiService.java
jhaeu: No, I removed many multiple blank lines, but maybe not all... I will reformat all changed classes.

ors-api/src/main/java/org/heigit/ors/api/services/AbstractApiService.java
jhaeu: changed

ors-api/src/test/java/org/heigit/ors/api/services/APIServiceTest.java
jhaeu: changed

ors-api/src/main/java/org/heigit/ors/api/services/AbstractApiService.java
jhaeu: Removed some double blank lines, but decided not to reformat the classes NOW - Amandus mentioned, that a big reformatting of the whole project in an extra branch is planned. BTW in our .editorconfig we seem to allow double blank lines if I'm not wrong?

takb (approved):
Looks all good to me now. Excellent! @aoles have your requests been addressed?

aoles (approved):
Thanks a lot @jhaeu for accommodating my suggestions, looks much cleaner now :)

I think we are good to go 🚀

Cheers!

PR #1506, opened 2023-07-24
feat: configuration using YAML files

### LINKED ISSUE #1499: Configuration Rework
### Problem description
The ors configuration is currently running with JSON and the code is having some ageing issues.

### Proposed solution

Streamline the configuration handling in code and move to .yaml configuration instead of JSON.
Remove the need to configure anything at all besides the most important parameters  (osm file etc) by setting more sensible defaults.

takb:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.
- [] 7. If changes/additions are made to the ors-config.json file, I have added these to the [ors config documentation][config]

- default configuration in application.yml
- apitests configuration in application-test.yml
- look for config files named `ors-config.yml in `/etc/openrouteservice/`, `~/.openrouteservice/` and `./` able to override the default settings
- deprecated AppConfig class and removed JSON files
- for the deprecation phase, old JSON configurations override every setting in yml files, temporary code bundled in `org.heigit.ors.api.util.AppConfigMigration` class
- some fixes to log formatting, log settings in JSON don't work anymore  
- refactoring of API request handling and limit checks


docker-entrypoint.sh
```diff
@@ -48,14 +44,5 @@ if [ ! -d "${catalina_base}/webapps/ors" ]; then
...
 fi
 
-if [ ! -f "$public_ors_config" ]; then
-  echo "No ors-config.json in ors-conf folder. Copying original config from ${source_ors_config}"
-  mkdir -p "${public_ors_config_folder}"
-  cp -f "${source_ors_config}" "${public_ors_config}"
-fi
-
-echo "Deploy ors with config from ${public_ors_config}"
-cp -f "${public_ors_config}" "${tomcat_ors_config}"
```
MichaelsJP: If this logic is gone now, we need to think how users could from now on configure their instances coherently across different deployments

MichaelsJP: Maybe instead of removing this logic, we just force copy the `ors-api/ors-config.yml` in that folder as e.g. `ors-config-example.ym` and always replace that with the current example config. That way, the users would stay up-to-date. If a `ors-config.yml` is in there, we could use that for configuring the instance.

MichaelsJP: What could be added in the future: Reading the individual config from env in the entrypoint.sh. We could introduce a namespace such as `ORS_CONFIG_XYZ` that could be set in .env files or in the docker-compose files and would be applied on startup.

ors-api/src/main/java/org/heigit/ors/api/EngineProperties.java
```diff
@@ -184,86 +361,149 @@ public void setExtStorages(Map<String, Map<String, String>> extStorages) {
...
+//            }
+//
+//            public int getMinOneWayNetworkSize() {
+//                return minOneWayNetworkSize;
+//            }
+//
+//            public void setMinOneWayNetworkSize(int minOneWayNetworkSize) {
+//                this.minOneWayNetworkSize = minOneWayNetworkSize;
+//            }
+//        }
```
MichaelsJP, line 508:
```suggestion
```
Old code?

ors-api/src/main/java/org/heigit/ors/api/EngineProperties.java
```diff
@@ -119,59 +185,170 @@ public String getCachePath() {
...
+            }
+            return output.toString();
+        }
+
         public void setEncoderOptions(Map<String, String> encoderOptions) {
             this.encoderOptions = encoderOptions;
         }
 
-        public PreparationProperties getPreparation() {
+//        public PreparationProperties getPreparation() {
```
jhaeu, line 325: remove commented out code?

docs/installation/Configuration.md
```diff
@@ -5,45 +5,111 @@ title: Configuration
...
-For now the easiest way to modify the settings described below. Note that all properties are translated from period-separated lowercase to underscore-separated uppercase, e.g. to override the `cors.allowed_origins` property you would set `CORS_ALLOWED_ORIGINS=https://example.org` (with `-e` flag to your `docker run` command or in the `environment` block in your `docker-compose.yml`). 
+The configuration of your own **openrouteservice** instance is done in a YAML file called `ors-config.yml`. *
+*openrouteservice** looks for such files in different locations. If a property is set in multiple files, later
+occurances will override previous ones.
 
-#### 2. Runtime parameters
-You can pass values to the JVM as java system variable. The example above could be set as `-Dcors.allowed_origins=https://example.org` (e.g. within the `JAVA_OPTS` variable through your `docker run` command or in the `environment` block in your `docker-compose.yml`)
+- `/etc/openrouteservice/ors-config.yml`
+- `~/.openrouteservice/ors-config.yml`
+- `./ors-config.yml` (in the runtime working directory)
```
MichaelsJP, line 15: As discussed earlier, this is ok for now. Ideally, ors should have a precedence:
- 1. Parameter for the jar start
- 2. ENV parameter for the config
- 3. A default location?

ors-api/pom.xml
```diff
@@ -246,6 +241,12 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-log4j2</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.logging.log4j</groupId>
+                    <artifactId>log4j-slf4j2-impl</artifactId>
+                </exclusion>
+            </exclusions>
```
MichaelsJP, line 249: Why the exclusion?

ors-api/src/main/java/org/heigit/ors/api/EngineProperties.java
```diff
@@ -184,86 +361,149 @@ public void setExtStorages(Map<String, Map<String, String>> extStorages) {
...
+
+        public String getGtfsFile() {
+            return gtfsFile != null ? gtfsFile : "";
+        }
+
+        public void setGtfsFile(String gtfsFile) {
+            this.gtfsFile = gtfsFile;
+        }
+
+//        public static class PreparationProperties {
```
jhaeu, line 478: remove comment?

ors-api/src/test/java/org/heigit/ors/apitests/common/InitializeGraphsOnce.java
```diff
@@ -23,21 +23,22 @@ public class InitializeGraphsOnce implements BeforeAllCallback, BeforeEachCallba
...
+            }
+        }
     }
 
     @Override
     public void beforeEach(ExtensionContext context) {
-        // Waiting for all graphs being built.
-        // Do it here - instead of beforeAll - because now the Logging configuration has been correctly set up.
-        RoutingProfileManager.getInstance();
     }
```
MichaelsJP: Remove empty override?

ors-api/src/main/java/org/heigit/ors/api/EngineProperties.java
takb: I'm hoping that we are going to refactor the `RoutingProfileConfiguration` after we have removed the old AppConfig. The commented out code is actually better but will make sense only after we have also refactored how these parameters are passed and used.

ors-api/src/main/java/org/heigit/ors/api/EngineProperties.java
takb: ditto

jhaeu (approved):
LGTM

.github/workflows/build-and-publish.yml
```diff
@@ -122,7 +122,7 @@ jobs:
           # Restart the container to apply the config changes
           docker stop ors-instance
           docker container prune -f 
-          docker run -it -d -p 8080:8080 -v $(pwd)/graphs:/home/ors/ors-core/data/graphs -v $(pwd)/conf:/home/ors/ors-conf -e CORS_ALLOWED_ORIGINS=https://example.org --name ors-instance ${{ needs.prepare_environment.outputs.test_image_name }}
+          docker run -it -d -p 8080:8080 -e ORS_CORS_ALLOWED_ORIGINS=https://example.org --name ors-instance ${{ needs.prepare_environment.outputs.test_image_name }}
```
MichaelsJP: Wouldn't it take longer without the mount?
From my understanding, the mount keeps the graph between the container runs and speeds up the second start.

Dockerfile
```diff
@@ -82,23 +65,24 @@ ENV LANG='en_US' LANGUAGE='en_US' LC_ALL='en_US'
 RUN apk add --no-cache bash=~'5' openssl=~'3' && \
     addgroup -g ${GID} ors && \
     adduser -D -h ${BASE_FOLDER} -u ${UID} -G ors ors &&  \
-    mkdir -p ${BASE_FOLDER}/ors-core/logs/ors ${BASE_FOLDER}/ors-conf ${BASE_FOLDER}/tomcat/logs &&  \
-    chown -R ors ${BASE_FOLDER}/tomcat ${BASE_FOLDER}/ors-core/logs/ors ${BASE_FOLDER}/ors-conf ${BASE_FOLDER}/tomcat/logs
+    mkdir -p ${BASE_FOLDER}/logs ${BASE_FOLDER}/graphs ${BASE_FOLDER}/elevation_cache ${BASE_FOLDER}/conf ${BASE_FOLDER}/data ${BASE_FOLDER}/tomcat/logs&&  \
+    chown -R ors ${BASE_FOLDER}/logs ${BASE_FOLDER}/graphs ${BASE_FOLDER}/elevation_cache ${BASE_FOLDER}/conf ${BASE_FOLDER}/data ${BASE_FOLDER}/tomcat ${BASE_FOLDER}/tomcat/logs
```
MichaelsJP: The last time we changed the folders, it caused a massive confusion for users. I'd say we keep it for 7.x and rewrite the paths for 8.
We could do it in one push by introducing jar and new paths.
What do you think?

docker-compose.yml
```diff
@@ -10,15 +10,14 @@ services:
...
-      - ./docker/elevation_cache:/home/ors/ors-core/data/elevation_cache
-      - ./docker/logs/ors:/home/ors/ors-core/logs/ors
+      - ./docker/graphs:/home/ors/graphs
+      - ./docker/elevation_cache:/home/ors/elevation_cache
+      - ./docker/logs/ors:/home/ors/logs
       - ./docker/logs/tomcat:/home/ors/tomcat/logs
-      - ./docker/conf:/home/ors/ors-conf
-      #- ./your_osm.pbf:/home/ors/ors-core/data/osm_file.pbf
+      - ./docker/conf:/home/ors/conf
+      - ./docker/data:/home/ors/data
```
MichaelsJP: Same as above

ors-api/src/main/java/org/heigit/ors/api/ORSEnvironmentPostProcessor.java
```diff
@@ -1,30 +1,46 @@
...
+            configLocations.add(System.getProperty(ORS_CONFIG_LOCATION_PROPERTY));
+        }
+        for (String path : configLocations) {
+            try {
+                List<PropertySource<?>> sources = this.loader.load("yml config", new FileSystemResource(path));
+                if (!sources.isEmpty()) {
+                    environment.getPropertySources().addAfter(StandardEnvironment.SYSTEM_ENVIRONMENT_PROPERTY_SOURCE_NAME, sources.get(0));
+                }
+            } catch (IllegalStateException | IOException ex) {
+                // Ignore yml file not present
```
MichaelsJP, line 42: This logic could youse some testing imo or at least some debug (not info) level logging, to know that's being hit easily.

ors-api/src/main/resources/log4j.properties
```diff
@@ -4,7 +4,7 @@ log4j.rootLogger=DEBUG, A1
 # A1 is set to be a ConsoleAppender.
 log4j.appender.A1=org.apache.log4j.ConsoleAppender
 log4j.appender.A1.layout=org.apache.log4j.PatternLayout
-log4j.appender.A1.layout.ConversionPattern=%d{dd MMM HH:mm:ss} %p [%c{2}] - %m%n
+#log4j.appender.A1.layout.ConversionPattern=%d{dd MMM HH:mm:ss} %p [%c{2}] - %m%n
```
MichaelsJP: Why not remove?

ors-engine/src/main/java/org/heigit/ors/routing/RoutingProfileManager.java
```diff
@@ -69,45 +68,60 @@ public static synchronized RoutingProfileManager getInstance() {
...
-                if (profilesTypes != null) {
+                if (rpc.getProfilesTypes() != null) {
                     Callable<RoutingProfile> task = new RoutingProfileLoader(config, rpc, loadCntx);
                     compService.submit(task);
                     nTotalTasks++;
                 }
             }
 
-            LOGGER.info("%d tasks submitted.".formatted(nTotalTasks));
+                LOGGER.info("%d tasks submitted.".formatted(nTotalTasks));
```
MichaelsJP:
```suggestion
            LOGGER.info("%d tasks submitted.".formatted(nTotalTasks));
```

MichaelsJP (changes requested):

docker-entrypoint.sh
takb: replaced with logic that copies over only if not present already, which should be enough to persist user made changes.

ors-api/src/main/java/org/heigit/ors/api/EngineProperties.java
takb: New code to be used after we refactor the parameter transport classes. Added a comment to mark as such.

ors-api/pom.xml
takb: Because another SLF4J provider is already pulled through dependencies elsewhere

docs/installation/Configuration.md
takb: program argument and env implemented, default location removed.

ors-api/src/test/java/org/heigit/ors/apitests/common/InitializeGraphsOnce.java
takb: Done.

.github/workflows/build-and-publish.yml
takb: Fixed.

Dockerfile
takb: Reverted to old messed up folder structure where possible. This might cause problems for user later because the old paths are now in the new yml configuration, and will require changing the config file once we fix it.

ors-api/src/main/java/org/heigit/ors/api/ORSEnvironmentPostProcessor.java
takb: At this point logging is not enabled yet, so it is not possible to output debug messages. I've added debug output in the InitContextListener instead that tells which configuration option was used.

ors-api/src/main/resources/log4j.properties
takb: Forgot to uncomment, I was just trying something. Fixed.

ors-engine/src/main/java/org/heigit/ors/routing/RoutingProfileManager.java
takb: Fixed.

docker-entrypoint.sh
```diff
@@ -42,20 +38,15 @@ if [ "${BUILD_GRAPHS}" = "True" ]; then
...
 
 echo "### openrouteservice configuration ###"
-if [ ! -d "${catalina_base}/webapps/ors" ]; then
-  echo "Extract war file to ${catalina_base}/webapps/ors"
-  cp -f "${ors_war_path}" "${catalina_base}"/webapps/ors.war
-  unzip -qq "${catalina_base}"/webapps/ors.war -d "${catalina_base}/webapps/ors"
+if [ ! -f "${ors_base}/ors-conf/ors-config.yml" ]; then
+  echo "Copy ors-config.yml"
+  cp -f "${ors_base}/tmp/ors-config.yml" "${ors_base}/ors-conf/ors-config.yml"
 fi
```
MichaelsJP, line 53:
```suggestion
# Always overwrite the example config in case another one is present
cp -f "${ors_base}/tmp/ors-config.yml" "${ors_base}/ors-conf/ors-config-example.yml"
# Check for old .json configs
JSON_FILES=$(ls -d -- ""${ors_base}/ors-conf/"*.json 2>/dev/null)
if [ -z "$JSON_FILES" ]; then
    echo "Old .json config found. They're depricated and will be replaced in ors version 8."
    echo "Please migrate to the new .yml example."
fi
# No config found. Use the base config
if [ ! -f "${ors_base}/ors-conf/ors-config.yml" ]; then
  echo "Copy ors-config.yml"
  cp -f "${ors_base}/tmp/ors-config.yml" "${ors_base}/ors-conf/ors-config.yml"
fi
```

MichaelsJP (changes requested):

PR #1508, opened 2023-07-26
perf(isochrones): Increase edge splitting threshold

aoles:
### Pull Request Checklist

- [ ] 1. I have [**rebased**][rebase] the latest version of the master branch into my feature branch and all conflicts
         have been resolved.
         [Unreleased] heading.

Fixes # .

- **Key functionality added:** 2x speedup of isochrone generation time.

- **Reason for change:** Setting the threshold for edge splitting to a low value slows down isochrone generation because of the increased amount of points which need to be processed when calculating the geometry. Initial benchmarks have revealed that increasing the current threshold of 20m to 200m yields acceptable results while drastically improving performance.

The edge splitting threshold has been verified by running some benchmark and assessing the results both in terms of performance and visual aspects of the generated geometries. All files are available on [GitLab](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/openrouteservice-infrastructure/benchmarks/-/tree/master/isochrones), while the resulting report can be viewed at [RPubs](https://rpubs.com/aoles/isochrones_edge_splitting).

The benchmark consisted of performing 10 isochrone queries for the car profile at 10 random locations within the DACH region. Each query requested a 60min isochrone split into 20min ranges. First, the cumulative distribution of edge distances have been gathered. This revealed that for the original threshold of 20m almost 70% of edges get split. Increasing this value 10-fold reduces the amount of splitted edges to 5% as illustrated by the following figure.
image

Then three scenarios have been evaluated: the original splitting of edges into 20m segments, splitting into 200m segments, and not doing any splitting at all. While the original 20m splitting introduces a significant 2-3x performance penalty, splitting based on 200m threshold is only 20% slower compared to no splitting. 
image

Not splitting long edges may occasionally introduce weird geometrical artifacts. The following figure compares an isochrone without edge splitting (red) to one calculated with 20m splitting.
image

These issues are mostly resolved with the 200m threshold while keeping the computation time low.

Blue is the original isochrone geometry while the red one was generated with this PR.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ RELEASING:
 - cleanup StatusCodeCaptureWrapper class ([#1504](https://github.com/GIScience/openrouteservice/pull/1504))
 - spring-boot-starter-parent to v3.1.1 ([#1504](https://github.com/GIScience/openrouteservice/pull/1504))
 - from springdoc-openapi-ui package to springdoc-openapi-starter-webmvc-ui ([#1504](https://github.com/GIScience/openrouteservice/pull/1504))
+- increase edge splitting threshold for generating isochrones ([#1508](https://github.com/GIScience/openrouteservice/pull/1508))
 
 ### Removed
 - dependency on apache-curator ([#1496](https://github.com/GIScience/openrouteservice/issues/1496))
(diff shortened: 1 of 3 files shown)


sfendrich (approved):
LGTM!

MichaelsJP:
@aoles Thanks for the detailed analysis on the effects of splitting. While not splitting imo is not acceptable, the balance between speed and accuracy with 200 m threshold looks good!

PR #1524, opened 2023-08-10
feat: add snapping endpoint

### LINKED ISSUE #1519: Integrate snapping endpoint
### Feature/Enhancement Proposal

Integrate the snapping enpoint proposed in https://ask.openrouteservice.org/t/obtain-snapped-location-data-outside-of-matrix/4835/2.

sfendrich:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes #1519 .

- Key functionality added: snapping endpoint that returns locations snapped to the nearest edge in the graph

### DIFF
--- a/ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
+++ b/ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
@@ -1,0 +1,163 @@
+/*
+ * This file is part of Openrouteservice.
+ *
+ * Openrouteservice is free software; you can redistribute it and/or modify it under the terms of the
+ * GNU Lesser General Public License as published by the Free Software Foundation; either version 2.1
+ * of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this library;
+ * if not, see <https://www.gnu.org/licenses/>.
+ */
+
+package org.heigit.ors.api.controllers;
+
+import com.fasterxml.jackson.databind.exc.InvalidDefinitionException;
+import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+import com.fasterxml.jackson.databind.exc.MismatchedInputException;
+import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import jakarta.servlet.http.HttpServletResponse;
+import org.heigit.ors.api.errors.CommonResponseEntityExceptionHandler;
+import org.heigit.ors.api.requests.snapping.SnappingApiRequest;
+import org.heigit.ors.api.responses.snapping.json.JsonSnappingResponse;
+import org.heigit.ors.api.services.SnappingService;
+import org.heigit.ors.exceptions.*;
+import org.heigit.ors.routing.APIEnums;
+import org.heigit.ors.snapping.SnappingErrorCodes;
+import org.heigit.ors.snapping.SnappingResult;
+import org.springframework.core.convert.ConversionFailedException;
+import org.springframework.http.ResponseEntity;
+import org.springframework.http.converter.HttpMessageConversionException;
+import org.springframework.http.converter.HttpMessageNotReadableException;
+import org.springframework.web.bind.MissingServletRequestParameterException;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+@Tag(name = "Snapping Service", description = "Snap coordinates to the graph edges.")
+@RequestMapping("/v2/snap")
+@ApiResponse(responseCode = "400", description = "The request is incorrect and therefore can not be processed.")
+@ApiResponse(responseCode = "404", description = "An element could not be found. If possible, a more detailed error code is provided.")
+@ApiResponse(responseCode = "405", description = "The specified HTTP method is not supported. For more details, refer to the EndPoint documentation.")
+@ApiResponse(responseCode = "413", description = "The request is larger than the server is able to process, the data provided in the request exceeds the capacity limit.")
+@ApiResponse(responseCode = "500", description = "An unexpected error was encountered and a more detailed error code is provided.")
+@ApiResponse(responseCode = "501", description = "Indicates that the server does not support the functionality needed to fulfill the request.")
+@ApiResponse(responseCode = "503", description = "The server is currently unavailable due to overload or maintenance.")
+public class SnappingAPI {
+    static final CommonResponseEntityExceptionHandler errorHandler = new CommonResponseEntityExceptionHandler(SnappingErrorCodes.BASE);
+
+    private final SnappingService snappingService;
+
+    public SnappingAPI(SnappingService snappingService) {
+        this.snappingService = snappingService;
+    }
+
+    // generic catch methods - when extra info is provided in the url, the other methods are accessed.
+    @GetMapping
+    @Operation(hidden = true)
+    public void getGetMapping() throws MissingParameterException {
+        throw new MissingParameterException(SnappingErrorCodes.MISSING_PARAMETER, "profile");
+    }
+
+    @PostMapping
+    @Operation(hidden = true)
+    public String getPostMapping(@RequestBody SnappingApiRequest request) throws MissingParameterException {
+        throw new MissingParameterException(SnappingErrorCodes.MISSING_PARAMETER, "profile");
+    }
+
+    // Matches any response type that has not been defined
+    @PostMapping(value = "/{profile}/*")
+    @Operation(hidden = true)
+    public void getInvalidResponseType() throws StatusCodeException {
+        throw new StatusCodeException(HttpServletResponse.SC_NOT_ACCEPTABLE, SnappingErrorCodes.UNSUPPORTED_EXPORT_FORMAT, "This response format is not supported");
+    }
+
+    // Functional request methods
+    @PostMapping(value = "/{profile}")
+    @Operation(
+            description = """
+                    Returns a list of points snapped to the nearest edge in the graph. In case an appropriate
+                    snapping point cannot be found within the specified search radius, "null" is returned.
+                    """,
+            summary = "Snapping Service"
+    )
+    @ApiResponse(
+            responseCode = "200",
+            description = "Standard response for successfully processed requests. Returns JSON.",
+            content = {@Content(
+                    mediaType = "application/json",
+                    schema = @Schema(implementation = JsonSnappingResponse.class)
+            )
+            })
+    public JsonSnappingResponse getDefault(@Parameter(description = "Specifies the route profile.", required = true, example = "driving-car") @PathVariable APIEnums.Profile profile,
+                                         @Parameter(description = "The request payload", required = true) @RequestBody SnappingApiRequest request) throws StatusCodeException {
+        return getJsonSnapping(profile, request);
+    }
+
+    @PostMapping(value = "/{profile}/json", produces = {"application/json;charset=UTF-8"})
+    @Operation(
+            description = """
+                    Returns a list of points snapped to the nearest edge in the graph. In case an appropriate
+                    snapping point cannot be found within the specified search radius, \"null\" is returned.
+                    """,
+            summary = "Snapping Service JSON"
+    )
+    @ApiResponse(
+            responseCode = "200",
+            description = "JSON Response.",
+            content = {@Content(
+                    mediaType = "application/json",
+                    schema = @Schema(implementation = JsonSnappingResponse.class)
+            )
+            })
+    public JsonSnappingResponse getJsonSnapping(
+            @Parameter(description = "Specifies the profile.", required = true, example = "driving-car") @PathVariable APIEnums.Profile profile,
+            @Parameter(description = "The request payload", required = true) @RequestBody SnappingApiRequest request) throws StatusCodeException {
+        request.setProfile(profile);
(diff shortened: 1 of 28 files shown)


ors-api/src/main/java/org/heigit/ors/api/responses/snapping/SnappingResponse.java
```diff
@@ -0,0 +1,12 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.heigit.ors.snapping.SnappingResult;
+public class SnappingResponse {
+    @JsonIgnore
+    private final SnappingResult result;
+
+    public SnappingResponse(SnappingResult result) {
+        this.result = result;
+    }
```
MichaelsJP, line 11: Is this already sufficient, in terms of swagger docs?

ors-api/src/main/java/org/heigit/ors/api/responses/snapping/json/JsonSnappingResponse.java
```diff
@@ -0,0 +1,37 @@
...
+import org.heigit.ors.api.responses.matrix.json.JSONIndividualMatrixResponse;
+import org.heigit.ors.api.responses.matrix.json.JSONLocation;
+import org.heigit.ors.api.responses.snapping.SnappingResponse;
+import org.heigit.ors.matrix.ResolvedLocation;
+import org.heigit.ors.snapping.SnappingResult;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Schema(name = "SnappingResponse", description = "The Snapping Response contains the snapped coordinates.")
```
MichaelsJP, line 18: Even though the response is quite simple it could make sense to have response examples for the docs api

ors-engine/src/main/java/org/heigit/ors/routing/APIEnums.java
```diff
@@ -172,6 +172,32 @@ public String toString() {
         }
     }
 
+    @Schema(name = "Snapping response type", description = "Format of the snapping response.")
```
MichaelsJP, line 175: @TheGreatRefrigerator Does the schema need more definition?

ors-api/src/test/java/org/heigit/ors/apitests/snapping/ParamsTest.java
```diff
@@ -0,0 +1,48 @@
...
+                .when()
+                .log().ifValidationFails()
+                .post(getEndPointPath() + "/{profile}/json")
+                .then()
+                .log().ifValidationFails()
+                .body("any { it.key == 'locations' }", is(true))
+                .statusCode(200);
+
+    }
+}
```
MichaelsJP: Imo a few more tests could be good.
At least one test checking failing requests.
Maybe two parametrized tests with different search radii could be better suited.

ors-engine/src/main/java/org/heigit/ors/snapping/SnappingRequest.java
```diff
@@ -0,0 +1,49 @@
...
+        String encoderName = RoutingProfileType.getEncoderName(profileType);
+        FlagEncoder flagEncoder = gh.getEncodingManager().getEncoder(encoderName);
+        PMap hintsMap = new PMap();
+        int weightingMethod = WeightingMethod.RECOMMENDED; // Only needed to create the profile string
+        ProfileTools.setWeightingMethod(hintsMap, weightingMethod, profileType, false);
+        ProfileTools.setWeighting(hintsMap, weightingMethod, profileType, false);
+        String CHProfileName = ProfileTools.makeProfileName(encoderName, hintsMap.getString("weighting", ""), false);
+        String profileName = CHProfileName;
+
+        // TODO: replace usage of matrix search context by snapping-specific class
```
MichaelsJP: What does the to do depend on, could this already be part of this PR?

MichaelsJP (dismissed):

TheGreatRefrigerator:
I can't run the normal Application run configuration of spring-boot.
It doesn't work with my old ors-config.json but even if i remove that and it uses the default files (after uncommenting the pbf line).
```
11 Aug 16:12:13 INFO                                                  main [ o.h.o.a.Application                      ]   Starting Application using Java 17.0.6 with PID 26175 (/Users/abutzer/Workspace/openrouteservice/ors-api/target/classes started by abutzer in /Users/abutzer/Workspace/openrouteservice)
11 Aug 16:12:13 INFO                                                  main [ o.h.o.a.Application                      ]   The following 1 profile is active: "default"
11 Aug 16:12:15 INFO                                              ORS-Init [ o.h.o.a.s.l.ORSInitContextListener       ]   Initializing ORS...
11 Aug 16:12:15 INFO                                              ORS-Init [ o.h.o.r.RoutingProfileManager            ]   Total - 110 MB, Free - 82.99 MB, Max: 4 GB, Used - 27.01 MB
11 Aug 16:12:15 ERROR                                             ORS-Init [ o.h.o.r.RoutingProfileManager            ]   
11 Aug 16:12:15 ERROR                                             ORS-Init [ o.h.o.r.RoutingProfileManager            ]   No profiles configured. Exiting.
11 Aug 16:12:15 ERROR                                             ORS-Init [ o.h.o.r.RoutingProfileManager            ]   
11 Aug 16:12:15 INFO                                                  main [ o.h.o.a.Application                      ]   Started Application in 2.803 seconds (process running for 3.698)
11 Aug 16:12:15 INFO                         SpringApplicationShutdownHook [ o.h.o.a.s.l.ORSInitContextListener       ]   Shutting down ORS and releasing resources.

```

ors-engine/src/main/java/org/heigit/ors/snapping/SnappingRequest.java
sfendrich: It mainly depends on priorities. Implementing these snapping-specific classes is only worth, if we do it cleanly. This would require significant effort with moderate impact. The list of more important cleanup issues is long.

ors-engine/src/main/java/org/heigit/ors/routing/APIEnums.java
TheGreatRefrigerator: If the enum is not used in the Body, this doesn't seem to be used for generating the OpenAPI specs. Same for the other enums in the file (RouteResponseType, MatrixResponseType, etc.).

So this is fine as is.

ors-api/src/main/java/org/heigit/ors/api/responses/snapping/SnappingResponse.java
TheGreatRefrigerator: Yes, the Response object types are built from the locations array and JSONLocation Object
grafik

ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
```diff
@@ -0,0 +1,157 @@
...
+    public void getInvalidResponseType() throws StatusCodeException {
+        throw new StatusCodeException(HttpServletResponse.SC_NOT_ACCEPTABLE, SnappingErrorCodes.UNSUPPORTED_EXPORT_FORMAT, "This response format is not supported");
+    }
+
+    // Functional request methods
+    @PostMapping(value = "/{profile}")
+    @Operation(
+            description = """
+                    Returns a list of points snapped to the nearest edge in the graph. \
+                    This method does not accept any request body or parameters other than profile, locations, maximum_snapping_radius.\
```
TheGreatRefrigerator: Is this line neccessary? Which parameters the endpoint accepts and the type is obvious with the swagger ui / API playground.

Probably replace with some info on Locations not found which result in `null`.

ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
```diff
@@ -0,0 +1,157 @@
...
+            )
+            })
+    public JsonSnappingResponse getDefault(@Parameter(description = "Specifies the route profile.", required = true, example = "driving-car") @PathVariable APIEnums.Profile profile,
+                                         @Parameter(description = "The request payload", required = true) @RequestBody SnappingApiRequest request) throws StatusCodeException {
+        return getJsonSnapping(profile, request);
+    }
+
+    @PostMapping(value = "/{profile}/json", produces = {"application/json;charset=UTF-8"})
+    @Operation(
+            description = "Returns a list of points snapped to the nearest edge in the graph.",
```
TheGreatRefrigerator: This should probably also contain the description of the other snapping endpoint, as they do the same.

Is there a reason for having always the `/json` endpoint as well?

ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
```diff
@@ -0,0 +1,157 @@
...
+                    Returns a list of points snapped to the nearest edge in the graph. \
+                    This method does not accept any request body or parameters other than profile, locations, maximum_snapping_radius.\
+                    """,
+            summary = "Snapping Service"
+    )
+    @ApiResponse(
+            responseCode = "200",
+            description = "Standard response for successfully processed requests. Returns JSON.",
+            content = {@Content(
+                    mediaType = "application/geo+json",
```
TheGreatRefrigerator: The response is not a geojson so this should be `application/json`. 
You can validate geojson with:
http://geojson.io/

See e.g. https://en.wikipedia.org/wiki/GeoJSON for geojson format.

Should we have that as response format as well?

ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
```diff
@@ -0,0 +1,157 @@
...
+    @PostMapping(value = "/{profile}/json", produces = {"application/json;charset=UTF-8"})
+    @Operation(
+            description = "Returns a list of points snapped to the nearest edge in the graph.",
+            summary = "Snapping Service JSON"
+    )
+    @ApiResponse(
+            responseCode = "200",
+            description = "JSON Response.",
+            content = {@Content(
+                    mediaType = "application/geo+json",
```
TheGreatRefrigerator: see above

ors-api/src/main/java/org/heigit/ors/api/responses/snapping/json/JsonSnappingResponse.java
TheGreatRefrigerator: I can see one example being generated in the swagger ui which is probably derived from the JSONLocation type:
grafik
This is ok i think.

ors-api/src/test/java/org/heigit/ors/apitests/snapping/ParamsTest.java
TheGreatRefrigerator: agree

TheGreatRefrigerator (changes requested):
One general question we should consider is wether we want to return this as geojson as well?

apart from that some smaller issues to resolve.

TheGreatRefrigerator:
@MichaelsJP @sfendrich Do we want this endpoint on the live API or should it be hidden in the Playground?
If yes we also need to consider this endpoint in the tyk policies and quota plans.

ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
sfendrich: Done.

ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
sfendrich: Done.

ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
sfendrich: Done.

ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
sfendrich: Done.

MichaelsJP:
> @MichaelsJP @sfendrich Do we want this endpoint on the live API or should it be hidden in the Playground? If yes we also need to consider this endpoint in the tyk policies and quota plans.

I'd say we open it up on the live API with the next major Release 8.

MichaelsJP:
> One general question we should consider is wether we want to return this as geojson as well?
> 
> apart from that some smaller issues to resolve.

@TheGreatRefrigerator Should be effortlessly integrable and would make the endpoint similar to the others. I'd say let's do it.

MichaelsJP:
So, just three things missing:
1. Decide on a request limit. We need one. If it's possible, we could just do it in the API-Gateway. It would make our lives easier.
2. Add a few more test cases (parametrized) and failing.
3. Add the GeoJSON return.
4. Test performance

Adjust the documentation?

ors-api/src/main/java/org/heigit/ors/api/controllers/SnappingAPI.java
TheGreatRefrigerator: the `/json` is there as we have the same structure in the other apis as well. Once we have `/geojson` it makes more sense again as well.

MichaelsJP:
> So, just three things missing:
> 
> 1. Decide on a request limit. We need one. If it's possible, we could just do it in the API-Gateway. It would make our lives easier.
We should integrate a config level request limit for the number of points to query. Even though we might add another check in the future in the api gateway

aoles (changes requested):
1. Add endpoint to `services` listed at localhost:8082/ors/v2/status.
2. Instead of the somewhat lengthy API parameter name `maximum_search_radius` wouldn't just `radius` be sufficient?

MichaelsJP:
Snap is added to Status

ors-engine/src/main/java/org/heigit/ors/snapping/SnappingRequest.java
aoles: At the time being the snap endpoint **works only if CH has been enabled** for a given profile and fails otherwise 🙈 This happens because of the dependency on the `routingCHGraph` in the following call.

https://github.com/GIScience/openrouteservice/blob/b392a0c5681fe73afa1ae1559bb6452598fba772/ors-engine/src/main/java/org/heigit/ors/snapping/SnappingRequest.java#L41

aoles (dismissed):
~The following TODO  needs to be addressed.~

https://github.com/GIScience/openrouteservice/blob/b392a0c5681fe73afa1ae1559bb6452598fba772/ors-engine/src/main/java/org/heigit/ors/snapping/SnappingRequest.java#L38

TheGreatRefrigerator (approved):
looks ready now

PR #1597, opened 2023-11-06
fix: revert rotated log files to being gzipped

takb:
Fixes removing `.gz` from rotated log files name pattern accidentally.


jhaeu (approved):
lgtm

PR #1623, opened 2023-12-06
feat: logging configuration

takb:
- Configure logging in one central YAML file
- Loglevel configuration in ors-config.yml
- Log file location and rotation can be configured using environment variables
- Renamed properties file for docker/tomcat deployment to mark as such (can be removed after switching to fat jar deployment)


sfendrich (approved):
LGTM!

PR #1648, opened 2024-01-16
feat: enhance output in case of wrong avoid areas

koebi:
This is an intermediate debugging release. Do not merge.
The only interesting change stems from the last commit.


ors-api/src/main/java/org/heigit/ors/api/requests/common/APIRequest.java
```diff
@@ -166,7 +191,7 @@ protected void validateAreaLimits(Polygon[] avoidAreas, int profileType) throws
                     }
                 }
             } catch (InternalServerException e) {
```
TheGreatRefrigerator: Is this Exception broad enough?

TheGreatRefrigerator (approved):
Looks good to me
Tested locally with spring-boot config. Exceptions work.

PR #1650, opened 2024-01-18
feat: mmap configuration

### LINKED ISSUE #1649: Make DataAccess properly configurable through YAML
GH fork can run with DataAccess classes configured to run in mmap mode, which uses less memory by not holding the entire graph data but loading only the parts immediately required. This also leads to a certain degree of performance loss, so this is currently an optional feature activated by configuration.

- [x] Introduce a proper config parameter
- [x] keep it behind a feature flag for now
   - it needs to be activated on demand for now.

takb:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes #1649  .

See Issue for details.

### DIFF
--- a/ors-api/ors-config.yml
+++ b/ors-api/ors-config.yml
@@ -12,6 +12,9 @@ ors:
 #    allowed_headers: Content-Type, X-Requested-With, accept, Origin, Access-Control-Request-Method, Access-Control-Request-Headers, Authorization
 #    preflight_max_age: 600
   engine:
+    ### use the following line to enable mmap mode
+#    graphs_data_access: MMAP
+
     ### use the following lines for development setup using test OSM file for Heidelberg
 #    source_file: ./src/test/files/heidelberg.osm.gz
 #    graphs_root_path: ./graphs
--- a/ors-api/ors-config.yml
+++ b/ors-api/ors-config.yml
@@ -12,17 +12,17 @@ ors:
 #    allowed_headers: Content-Type, X-Requested-With, accept, Origin, Access-Control-Request-Method, Access-Control-Request-Headers, Authorization
 #    preflight_max_age: 600
   engine:
+    graphs_root_path: ./ors-core/data/graphs
+
     ### use the following line to enable mmap mode
 #    graphs_data_access: MMAP
 
     ### use the following lines for development setup using test OSM file for Heidelberg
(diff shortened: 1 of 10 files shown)


ors-api/ors-config.yml
```diff
@@ -7,13 +7,20 @@
...
+#    allowed_headers: Content-Type, X-Requested-With, accept, Origin, Access-Control-Request-Method, Access-Control-Request-Headers, Authorization
+#    preflight_max_age: 600
   engine:
-    ### use the following line for development setup using test OSM file for Heidelberg
-    #source_file: ./src/test/files/heidelberg.osm.gz
+    ### use the following line to enable mmap mode
+#    graphs_data_access: MMAP
+
+    ### use the following lines for development setup using test OSM file for Heidelberg
+#    source_file: ./src/test/files/heidelberg.osm.gz
```
aoles: Does the path rooted in `./src/...` make sense? It is inconsistent with the uncommented part which uses `./ors-core`. I think the recommendation is to have the working directory set to the project parent dir `openrouteservice`, at least this is how I have it set up locally..

aoles (changes requested):
Thanks! Looks quite good, see my comment for some minor issue I've noticed.

ors-api/ors-config.yml
takb: Okay, I'll need to make sure the paths in the docs are good

ors-api/ors-config.yml
```diff
@@ -12,17 +12,17 @@ ors:
 #    allowed_headers: Content-Type, X-Requested-With, accept, Origin, Access-Control-Request-Method, Access-Control-Request-Headers, Authorization
 #    preflight_max_age: 600
   engine:
+    graphs_root_path: ./ors-core/data/graphs
```
aoles, line 15: Is there any specific reason why the default is to have `ors-core` subdirectory? For example, when ORS is run from the command line the cwd is set to `ors-api`, so the paths become `ors-api/./ors-core/...`. I would suggest to skip on `ors-core` entirely.

takb: Yes, it reflects the setup used within the docker container. Personally I think that's something we could get rid of when we redo the docker setup after we merge the jar deployment...

aoles (approved):
Cheers! 🚀

PR #1607, opened 2023-11-14
perf(isochrones): improve isochrones performance

rtroilo:
Replacing `ConcaveHullOpenSphere` with jts own `ConcaveHull` algorithm

### Pull Request Checklist

- [x] 1. I have [**rebased**][rebase] the latest version of the master branch into my feature branch and all conflicts
         have been resolved.
- [ ] 2. I have added information about the change/addition to functionality to the CHANGELOG.md file under the
         [Unreleased] heading.
- [ ] 3. I have documented my code using JDocs tags.
- [ ] 4. I have removed unnecessary commented out code, imports and System.out.println statements.
- [ ] 5. I have written JUnit tests for any new methods/classes and ensured that they pass.
- [ ] 6. I have created API tests for any new functionality exposed to the API.
- [ ] 7. If changes/additions are made to the ors-config.json file, I have added these to the [ors config documentation][config]
         along with a short description of what it is for, and documented this in the Pull Request (below).
- [x] 8. I have built graphs with my code of the Heidelberg.osm.gz file and run the api-tests with all test passing
- [x] 9. I have referenced the Issue Number in the Pull Request (if the changes were from an issue).
- [x] 10. For new features or changes involving building of graphs, I have tested on a larger dataset
          (at least Germany), and the graphs build without problems (i.e. no out-of-memory errors).
- [ ] 11. For new features or changes involving the graphbuilding process (i.e. changing encoders, updating the
          importer etc.), I have generated longer distance routes for the affected profiles with different options
          (avoid features, max weight etc.) and compared these with the routes of the same parameters and start/end
          points generated from the current live ORS.
          If there are differences then the reasoning for these **MUST** be documented in the pull request.
- [x] 12. I have written in the Pull Request information about the changes made including their intended usage
          and why the change was needed.
- [ ] 13. For changes touching the API documentation, I have tested that the API playground [renders correctly][api].

Fixes #1573.

### Information about the changes
- Key functionality added:
- Reason for change:

### Examples and reasons for differences between live ORS routes, and those generated from this pull request
-

### Required changes to ors config (if applicable)
-

[config]: https://GIScience.github.io/openrouteservice/installation/Configuration.html
[api]: https://gitlab.gistools.geog.uni-heidelberg.de/giscience/openrouteservice-infrastructure/ors-docs-api#test-new-ors-documentation
[rebase]: https://github.com/GIScience/openrouteservice/blob/master/CONTRIBUTE.md#pull-request-guidelines

PR #1658, opened 2024-01-30
perf(fastisochrones): apply modifications from regular isochrones

aoles:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Build- and query-time optimizations based on the previous work on regular isochrones in #1607:
* Replace concave hull algorithm with JTS implementation
* Do not perform neighbor search when adding new elements to the point cloud, both for preparation and for query
* Reduce splitting length according to splitting factor
* Avoid unnecessary edge splitting in  fully reachable and active cells

Changes to isochrone geometries in API tests (**blue** is reference, **red** is new)

#### testIntersections

image

#### testPolygon

image

#### testSmoothing

image
image


aoles:
Comprehensive assessment of the changes introduced by this PR:
* [Performance benchmark](https://rpubs.com/aoles/fast_isochrones_performance_benchmarks)
* [Modifications to geometries](https://rpubs.com/aoles/fast_isochrones_geometries)
* Detailed analysis of geometry modifications based on notorious [location 7](https://rpubs.com/aoles/fast_isochrones_example)

To summarize, changes introduced to the **preparation part** (constructing cell geometries) doesn't seem to have much impact on the overall performance, neither the actual [build times](https://heigit.atlassian.net/wiki/x/lID1Ew), nor the query times. Some small differences in the resulting geometries are observed after switching the algorithm from `ConcaveHullOpenSphere` to JTS `ConcaveHull`. The actual performance gain comes from adapting the **query time optimizations**:
* removing neighbor search
* and reducing the splitting length based on the smoothing factor

each of which individually improves the performance by around 18% and 17%, respectively. **The combined performance gain introduced by this PR is around 33%**.

Last but not least, there now seem to be a better overlap with the geometries produced by regular isochrones.

CHANGELOG.md
```diff
@@ -72,7 +72,7 @@ RELEASING:
 - fix IN1-JAVA-ORGMOZILLA-1314295 ([#1627](https://github.com/GIScience/openrouteservice/issues/1627))
 - log summary stats on traffic mapmatching and use progress bar only in debug mode ([#1647](https://github.com/GIScience/openrouteservice/pull/1647))
 - move APIEnums into API module ([#1634](https://github.com/GIScience/openrouteservice/issues/1634))
-- performance improvements to isochrone calculations ([#1607](https://github.com/GIScience/openrouteservice/pull/1607))
+- performance improvements to isochrone calculations ([#1607](https://github.com/GIScience/openrouteservice/pull/1607)) and ([#1658](https://gith$
```
sfendrich: It looks like the end of the line has been cut off.

ors-engine/src/main/java/org/heigit/ors/isochrones/builders/fast/FastIsochroneMapBuilder.java
```diff
@@ -343,7 +335,7 @@ private List<Double> createCoordinateListFromGeometry(Geometry preprocessedGeome
                             ring.getPointN(i).getX(),
                             ring.getPointN(i + 1).getX(),
                             contourCoordinates,
-                            MIN_EDGE_LENGTH_LIMIT,
+                            minSplitLength/2,
```
sfendrich, line 338: Why is this divided by two, and should we then directly pass the value already divided?

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
```diff
@@ -414,8 +414,7 @@ protected void postProcessing(boolean closeEarly) {
                 calculateContours();
                 List<Profile> profiles = fastIsochroneFactory.getFastIsochroneProfiles();
                 for (Profile profile : profiles) {
-                    Weighting weighting = ((ORSWeightingFactory) createWeightingFactory()).createIsochroneWeighting(profile, new PMap(profile.getName()).putObject("isochroneWeighting", "true"));
-
+                    Weighting weighting = ((ORSWeightingFactory) createWeightingFactory()).createIsochroneWeighting(profile);
```
sfendrich, line 417: I wonder whether there was a reason to have a seperate `IsochroneWeighting`?

sfendrich (changes requested):
Just a few minor questions.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSGraphHopper.java
aoles: Do you mean the removed PMap entry `"isochroneWeighting"`? I couldn't find any use of it, so to me it looks like a historic thing which became obsolete at some point...

ors-engine/src/main/java/org/heigit/ors/isochrones/builders/fast/FastIsochroneMapBuilder.java
aoles: This is so because the initial hope was that `minSplitLength` would be enouggh, however `minSplitLength`is is the correct value only for edges around the equator. I figured out later that the whole approach of splitting based on the length in meters is incompatible with how the concave hull is actually being computed in the geocoordinate domain. The current solution of taking half of the length is more of a heuristic which seems to work pretty well, a proper fix is underway in a separate PR 🤞

CHANGELOG.md
aoles: Thanks, sorry for that! 🙈 I've fixed it.

sfendrich (approved):
LGTM! Thank you!

PR #1666, opened 2024-01-31
fix: no config loading if test profile is active

takb:
- Reason for change:

Added a check in the ORSEnvironmentPostProcessor to skip configuration files loading if the "test" profile is active. That way the API tests won't break in the future because of changes to `ors-config.yml`


MichaelsJP (approved):
LGTM

PR #1670, opened 2024-01-31
feat(docker): Introduce reworked docker setup with jar

### LINKED ISSUE #1664: Move to JAR deployment in docker
https://github.com/GIScience/openrouteservice/pull/1507

MichaelsJP:
This PR introduces a completely overhauled docker setup. Due to the nature of the changes, the old docker setup isn't compatible with this one. No new major incompatibilities are to be expected soon.

By moving away from the old (almost 6-year-old docker structure) we now have the following features available:
- Coherent user permissions
- Clear mount points
- Well documented docker-compose.yml
- Reworked entrypoint with log levels, sensible log messages, debug mode config less configuration etc.
- Reworked dockerfile that is roughly 1/3 in terms of code lines and a lot less complex.
- We now use the jar file in the container. No more .war-File back and forth (finally).
- Multiple ways of configuring ors: config files (yml, deprecated json), container env variables á la `ors.*.someting`
- **Cloud-ready image**. Te image can be 100% configured by environment variables (env files) and therefore run in any cloud and Kubernetes infrastructure without much hustle.
- The image itself can be used as a base for custom cloud images as well.

### Pull Request Checklist

- [ ] 1. I have [**rebased**][rebase] the latest version of the main branch into my feature branch and all conflicts
         have been resolved.
         [Unreleased] heading.

Fixes # .

- Key functionality added:
- Reason for change:

-

-


TheGreatRefrigerator:
rebased on main

TheGreatRefrigerator:
rebased on main and fixed conflicts

TheGreatRefrigerator:
Currenlty includes the nightly build commit
- Restructure commits
- rework conversion & extract validation
- resolve todos (link to docu & add default env properties for java opts)

TheGreatRefrigerator:
rebased on main

TheGreatRefrigerator:
The test runs still fail due to the new setup prefering example-ors-config.yml over ors-config.yml.

Here the docs from my previous investigation
[docker_run_and_compose.md](https://github.com/GIScience/openrouteservice/files/14497520/docker_run_and_compose.md)

Dockerfile
```diff
@@ -1,98 +1,61 @@
...
 
 # Setup the target system with the right user and folders.
-RUN apk add --no-cache bash=~'5' openssl=~'3' && \
-    addgroup -g ${GID} ors && \
-    adduser -D -h ${BASE_FOLDER} -u ${UID} -G ors ors &&  \
-    mkdir -p ${BASE_FOLDER}/ors-core/logs ${BASE_FOLDER}/ors-conf ${BASE_FOLDER}/ors-core/data ${BASE_FOLDER}/tomcat/logs &&  \
-    chown -R ors ${BASE_FOLDER}/tomcat ${BASE_FOLDER}/ors-core/logs ${BASE_FOLDER}/ors-conf ${BASE_FOLDER}/ors-core/data ${BASE_FOLDER}/tomcat/logs
-
-WORKDIR ${BASE_FOLDER}
+RUN apk update && apk add --no-cache bash openssl yq jq curl && \
```
koebi: why do we install openssl and curl here?

docker-entrypoint.sh
```diff
@@ -1,61 +1,366 @@
...
+
+echo "###########################"
+echo "# Container sanity checks #"
+echo "###########################"
+info "Running container as user $(whoami) with id $(id -u) and group $(id -g)"
+if [[ $(id -u) -eq 0 ]] || [[ $(id -g) -eq 0 ]] ; then
+  debug "User and group are set to root with id 0 and group 0."
+elif [[ $(id -u) -eq 1000 ]] || [[ $(id -g) -eq 1000 ]] ; then
+  debug "User and group are set to 1000 and group 1000."
+else
```
koebi, line 141: Given the line above, this seems redundant?

docker-entrypoint.sh
```diff
@@ -1,61 +1,366 @@
...
 
-ors_base=${1}
-catalina_base=${ors_base}/tomcat
-graphs=${ors_base}/ors-core/data/graphs
+success "ORS_HOME: ${ORS_HOME} exists and is writable."
 
-echo "ORS Path: ${ors_base}"
-echo "Catalina Path: ${catalina_base}"
+mkdir -p "${ORS_HOME}"/{files,logs,graphs,elevation_cache,config} || warning "Could not create ${ORS_HOME} and folders"
+mkdir -p "${ORS_HOME}"/{files,logs,graphs,elevation_cache,config} || warning "Could not create ${ORS_HOME} and folders"
```
koebi: `ORS_HOME` is created above, so this should only report errors on folders.
Also, why is this done twice?

docker-entrypoint.sh
```diff
@@ -1,61 +1,366 @@
...
+# get ors.engine.graphs_root_path=. Dot notations in bash are not allowed, so we need to use awk to parse it.
+ors_engine_graphs_root_path=$(env | grep "^ors\.engine\.graphs_root_path=" | awk -F '=' '{print $2}')
+# get ors.engine.elevation.cache_path
+ors_engine_elevation_cache_path=$(env | grep "^ors\.engine\.elevation\.cache_path=" | awk -F '=' '{print $2}')
+# get ors.engine.source_file
+ors_engine_source_file=$(env | grep "^ors\.engine\.source_file=" | awk -F '=' '{print $2}')
+# get ors_config_location
+ors_config_location=${ORS_CONFIG_LOCATION:-""}
+# Unset the ORS_CONFIG_LOCATION to not interfere with the spring-boot application
+unset ORS_CONFIG_LOCATION
```
koebi: This is only used in line 217, so we should move it there?

docker-entrypoint.sh
```diff
@@ -1,61 +1,366 @@
...
+  ors_rebuild_graphs="false"
+fi
+
+
+# get ors.engine.graphs_root_path=. Dot notations in bash are not allowed, so we need to use awk to parse it.
+ors_engine_graphs_root_path=$(env | grep "^ors\.engine\.graphs_root_path=" | awk -F '=' '{print $2}')
+# get ors.engine.elevation.cache_path
+ors_engine_elevation_cache_path=$(env | grep "^ors\.engine\.elevation\.cache_path=" | awk -F '=' '{print $2}')
+# get ors.engine.source_file
+ors_engine_source_file=$(env | grep "^ors\.engine\.source_file=" | awk -F '=' '{print $2}')
```
koebi: These are only used in lines 234 onwards, so we should move it to there?

docker-entrypoint.sh
```diff
@@ -1,61 +1,366 @@
...
-if [ "${BUILD_GRAPHS}" = "True" ]; then
-  rm -rf "${graphs:?}"/*
+# The config situation is difficult due to the recent ors versions.
+# To ensure a smooth transition, we need to check if the user is using a .json file or a .yml file.
+# If neither is set, we need to print a migration info and default to the example-ors-config.env file.
+# Check if ors_config_location is a .json file and exists
+if [[ "${ors_config_location}" = *.yml ]] && [[ -f "${ors_config_location}" ]]; then
+  success "Using yml config: ${ors_config_location}"
+elif [[ "${ors_config_location}" = *.json ]] && [[ -f "${ors_config_location}" ]]; then
+  success "Using json config: ${ors_config_location}"
```
koebi: Here, we could note that we're using the values from the ENV?

docker-entrypoint.sh
```diff
@@ -1,61 +1,366 @@
...
 
-if [ ! -f "${ors_base}/ors-core/data/osm_file.pbf" ]; then
-  echo "Copy osm_file.pbf"
-  cp -f "${ors_base}/tmp/osm_file.pbf" "${ors_base}/ors-core/data/osm_file.pbf"
+if [[ -z "${ors_engine_source_file}" ]] && [[ "${ors_config_location}" = *.yml ]]; then
+  ors_engine_source_file=$(extract_config_info "${ors_config_location}" '.ors.engine.source_file')
+fi
+
+if [[ -z "${ors_engine_source_file}" ]] && [[ "${ors_config_location}" = *.json ]]; then
+  ors_engine_source_file=$(extract_config_info "${ors_config_location}" '.ors.services.routing.sources[0]')
```
koebi: We could refactor this to reflect the:
```
does it exist?
   either get from yml
   or get from json
```
structure

koebi (changes requested):
I do have a few comments on Dockerfile/compose/entrypoint.
I'll look at the utils and workflows independently.

Dockerfile
MichaelsJP: Good question regarding OpenSSL and curl. I think remnants from earlier dev steps. Will remove them.

docker-entrypoint.sh
MichaelsJP: The goal here is more to react with an explicit debug message on both scenarios individually. Either 0 or 1000 have different implications for the runtime container.

docker-entrypoint.sh
MichaelsJP: adjusted

docker-entrypoint.sh
MichaelsJP: I think it's better to have related functionality in one place. The initialization of the container env is done on that section.

docker-entrypoint.sh
MichaelsJP: same as above. i think the init of the container env should be done in one place completely.

docker-compose.yml
```diff
@@ -1,26 +1,84 @@
...
+    # Values will be overwritten if set in the 'environment' section.
+    #env_file: ors-config.env
+
+    # ----------------- Healthcheck configuration ------------------- #
+    # The healthcheck is disabled by default. Uncomment the following lines to enable it.
+    # The healthcheck allows you to monitor the status of the ORS application.
+    # Be careful with long graph builds, as the healthcheck will fail and show 'unhealthy' during this time.
+    # This is especially useful if you run your ORS container in a production environment.
+    #healthcheck:
+    #  test: curl --fail http://localhost:8082/ors/v2/health || exit 1
```
koebi: I just found that this is where curl is needed. I think we should switch to wget here.

docker-entrypoint.sh
MichaelsJP: yep

MichaelsJP: what about the following line after this if else construct?
```
info "Any ENV variables will have precedence over configuration variables from config files."
```

docker-entrypoint.sh
MichaelsJP: @jhaeu please revisit this section. You've reworked this quite a lot.

docker-compose.yml
MichaelsJP: why wget?

MichaelsJP: and yes. that was the reason. good catch

koebi: wget should be available by default, so we don't need to install anything else

docker-entrypoint.sh
koebi: Yes, but since these variables are only used there and possibly even overwritten, I'd say the belong to that section, and shouldn't be declared way before, so no one remembers what they even were?

docker-entrypoint.sh
koebi: That is valid information for the user, but these two lines are only printed if the value is read from an ENV. So I'm thinking along the lines of "Using yml config from ENV" or similar?

docker-entrypoint.sh
jhaeu: I think the block I changed was the previous one. 
Pushed a refactored code block respecting @koebi 's suggestion (hope I got it right)

docker-entrypoint.sh
MichaelsJP: Well, it's a matter of taste.  If you think it brings more value we can do that.
Fine for me :+1:

docker-entrypoint.sh
MichaelsJP: Well, it's a matter of taste.  If you think it brings more value we can do that.
Fine for me :+1:

docker-entrypoint.sh
MichaelsJP: I'll change it to the following:
```
success "Using yml config from ENV: ${ors_config_location}"
success "Using json config from ENV: ${ors_config_location}"
```

docker-compose.yml
MichaelsJP: gotya

docker-entrypoint.sh
koebi: I think it makes the code clearer and easier to read, yes.

docker-entrypoint.sh
koebi: Can these implications be easily mentioned? If so, we might want to do that.
If not, that's fine by me.

koebi (changes requested):

docker-entrypoint.sh
MichaelsJP: btw. they're used in line 183 the first time. so i'll move them there

docker-entrypoint.sh
MichaelsJP: Nope. It depends on how they read the documentation and what way they set it up. When people temper with permissions, the container needs to feed back what it is ultimately using. Else it'll become a black box in that regard.

MichaelsJP:
@koebi Cannot answer on your comments directly, so I'll answer it here:

> I think it makes the code clearer and easier to read, yes.

Fixed. I moved them to its first usage in line 182.

> Can these implications be easily mentioned? If so, we might want to do that. If not, that's fine by me.

Not really. It depends on how they read the documentation and what way they're configuring the container user. 
Based on that, a simple feedback from the image on the internal is useful. Else, the container will become a small black box.

koebi (approved):

PR #1679, opened 2024-02-05
refactor: remove logging configuration

takb:
- Reason for change:
Logging used to have a separate log4j configuration file, we get rid of those since using the standard spring logging configuration instead should suffice for our needs. 
Also removing an old log4j configuration used for the Tomcat in the  docker setup.


MichaelsJP (approved):
LGTM

PR #1682, opened 2024-02-07
fix(isochrones): avoid speed reduction on ways tagged with `access=destination`

aoles:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes # .

- Key functionality added: ignore speed penalty on roads tagged with `access = destination` when computing isochrones.
- Reason for change: isochrones starting around ways tagged as `access = destination` were smaller than expected

### Examples and reasons for differences between live ORS routes, and those generated from this pull request.

Original 5-minutes isochrone (the smaller inner polygon) vs. one generated with this PR (the larger enclosing polygon).

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@ RELEASING:
 - log summary stats on traffic mapmatching and use progress bar only in debug mode ([#1647](https://github.com/GIScience/openrouteservice/pull/1647))
 - move APIEnums into API module ([#1634](https://github.com/GIScience/openrouteservice/issues/1634))
 - performance improvements to isochrone calculations ([#1607](https://github.com/GIScience/openrouteservice/pull/1607))
+- do not apply speed penalty to roads tagged with "access=destination" when computing isochrones ([#1682](https://github.com/GIScience/openrouteservice/pull/1682))
 
 ### Deprecated
 - JSON configuration and related classes ([#1506](https://github.com/GIScience/openrouteservice/pull/1506))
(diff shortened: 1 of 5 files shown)


koebi:
Hey,

two questions:
1. Can you explain what is happening and why it is solving the issue?
2. This is another workaround for the workaround that `access:destination` is penalized instead of handled correctly, right? Do we have an issue for that?

sfendrich (approved):
LGTM!

aoles:
Thanks @koebi !

A proper solution for handling `access:destination` tags via a priority-type of weighting has been already implemented in https://github.com/GIScience/openrouteservice/pull/1412. This solution uses `LimitedAccessWeighting` to wrap around a given weighting returned by https://github.com/GIScience/openrouteservice/blob/20ddcbf5c420a702f992a7126d03b116d0f9af7c/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/ORSWeightingFactory.java#L55

The weighting for isochrones, however, is initialized outside of this method and was originally set to either GH's `FastestWeighting` or `ShortestWeighting`. The problem with `FastestWeighting` is that GH has their own, speed-based solution to penalize `access:destination` ways. Therefore, `ORSFastestWeighting` which overrides and disables this behavior should be used instead.

One could argue that for isochrone calculation `LimitedAccessWeighting` should be used as well. However, in the current implementation this won't work because the numbers returned by the weighting are interpreted as time counted to the isochrone limit, which effectively acts like a speed limit. A proper solution would require keeping separately track of the weighting used by Dijkstra, and the actually elapsed travel time for the stopping criterion 🤓

koebi:
But `LimitedAccessWeighting` also penalizes `access:destination` instead of blocking access if neither start nor end nor any waypoint are inside the `access:destination` area, as the restriction is intended, right?

PR #1687, opened 2024-02-13
fix: only run action if config changed

### LINKED ISSUE #1685: Simplify checking if config has changed
The workflow currently always checks out the repo and does some stuff before checking if the config has changed:
https://github.com/GIScience/openrouteservice/blob/e7b3bad932d504c7c72c83983caf129d4ad4f349/.github/workflows/config-conversion-automation.yml#L38

This could be done quicker and easier via the [`paths` filter for actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore).

koebi:
Using github.event.pull_request and the paths-filter, we can do the same but shorter.

fixes #1685

### DIFF
--- a/.github/workflows/config-conversion-automation.yml
+++ b/.github/workflows/config-conversion-automation.yml
@@ -1,56 +1,29 @@
-name: Automatic conversion of application.yml to ors-config.yml
+name: Automatic synchronization of application.yml to ors-config.yml
 on:
   pull_request:
     branches:
       - main
       - releases/**
-    types: [ edited, opened, ready_for_review, review_requested, reopened, synchronize ]
-  workflow_dispatch:
-
+    types: [ opened, ready_for_review, reopened, synchronize ]
+    paths:
+      - 'ors-api/src/main/resources/application.yml'
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/config-conversion-automation.yml
```diff
@@ -4,7 +4,9 @@ on:
     branches:
       - main
       - releases/**
-    types: [ edited, opened, ready_for_review, review_requested, reopened, synchronize ]
+    types: [ opened, ready_for_review, reopened, synchronize ]
+    paths:
+      - 'ors-api/scr/main/resources/application.yml'
```
TheGreatRefrigerator: wrong path `scr`

TheGreatRefrigerator (changes requested):

TheGreatRefrigerator (approved):

.github/workflows/config-conversion-automation.yml
TheGreatRefrigerator: Done

PR #1689, opened 2024-02-14
ci(docker-nightly): move to separate build/publish from test

### LINKED ISSUE #1686: fix docker nightly build

- [X] I have searched the existing issues

### Problem description

Currently, docker nightly images get built & published to dockerhub for basically every PR change, which consumes quite some quota.
This is neither a nightly build nor is it consistent, as the content of the image changes with the branch currently being built.

### Proposed solution

Do normal nightly build if anything changed in `main`.
Similar to e.g. https://github.com/libgit2/libgit2/blob/main/.github/workflows/nightly.yml

### Additional context

If we still want to do docker image builds for PRs we should do it differently

### Forum Topic Link

_No response_

TheGreatRefrigerator:
previously the nightly build was not built nightly but on many changes in PRs as well.
This is now moved to a separate workflow that runs nightly, with workflow_dispatch option to build images for different branches & repos on demand.

### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes https://github.com/GIScience/openrouteservice/issues/1686

- Key functionality added: real nightly build
- Reason for change: nightly was not built nightly

-

-

### DIFF
--- a/.github/workflows/docker-nightly-image.yml
+++ b/.github/workflows/docker-nightly-image.yml
@@ -1,0 +1,87 @@
+name: Docker Nightly Image CI
+run-name: Build ${{ inputs.branch }} - ${{ inputs.user }}
+
+on:
+  schedule:
+    - cron: '15 1 * * *'
+  workflow_dispatch:
+    inputs:
+      repository:
+        description: 'GitHub repository to create image off.'
+        required: true
+        default: 'GIScience/openrouteservice'
+      branch:
+        description: 'GitHub branch to create image off.'
+        required: true
+        default: 'main'
+      tag:
+        description: 'Name of the docker tag to create.'
+        required: true
+        default: 'nightly'
+      ignore-24h-commit-check:
+        description: 'Build image regardless of last commit date.'
+        required: true
+        type: boolean
+        default: false
+      user:
+        description: ''
+        required: false
+        default: 'schedule'
+
+jobs:
+  check-commits:
+    runs-on: ubuntu-latest
+    outputs:
+      new_commits: ${{ steps.check.outputs.NEW_COMMITS }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4.1.1
+        with:
+          repository: ${{ inputs.repository || 'GIScience/openrouteservice' }}
+          ref: ${{ inputs.branch || 'main' }}
+
+      - name: Check for new commits since 24 h ago
+        id: check
+        run: |
+          export LOG_LINES=$(git log --oneline --since '24 hours ago')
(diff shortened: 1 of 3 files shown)


TheGreatRefrigerator:
The check for new commits doesn't work properly yet

.github/workflows/docker-nightly-image.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Build and push
+        uses: docker/build-push-action@v5.1.0
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64/v8
+          provenance: false
+          push: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
MichaelsJP, line 86: `Push` true and `cache-to` together doesn't work, if I recall it correctly.

TheGreatRefrigerator: do you have have more details on this?
as far as i understand the action docs that link to the docker [cache-to](https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-to) for this and those examples there:
https://github.com/moby/buildkit?tab=readme-ov-file#github-actions-cache-experimental

it should work together?

MichaelsJP: Hm. When I look it up, I cannot find it any more. I remember it in earlier versions being absolutely annoying about some combinations of parameters. I guess you tested it, so let's leave it if the action isn't complaining :)!

.github/workflows/docker-nightly-image.yml
```diff
@@ -0,0 +1,87 @@
...
+        uses: actions/checkout@v4.1.1
+        with:
+          repository: ${{ inputs.repository || 'GIScience/openrouteservice' }}
+          ref: ${{ inputs.branch || 'main' }}
+
+      - name: Check for new commits since 24 h ago
+        id: check
+        run: |
+          export LOG_LINES=$(git log --oneline --since '24 hours ago')
+          export NUM_LINES=$(echo "$LOG_LINES" | grep -c .)
```
jhaeu, line 47:
```suggestion
          export NUM_LINES=$(echo "$LOG_LINES" | wc -l)
```

jhaeu (approved):
LGTM, wc -l is just a suggesten

.github/workflows/docker-nightly-image.yml
TheGreatRefrigerator: wc -l i tried, it gives indentation as output (`           0` instead of `0`), so the check for o 0/1 will be difficult.
That's why i switched to grep

PR #1698, opened 2024-02-22
build(deps): Cleanup maven dependencies

sfendrich:
- [x] declare undeclared used dependencies
- [x] remove unused declared dependencies

This PR cleans up the maven dependencies by declaring undeclared used dependencies and by commenting out unused declared dependencies as identified by `mvn dependency:analyze`. The unused dependencies have been commented out one by one with all unit and integration tests passing. However, there might still be some cases not covered by tests that need some of the dependencies at runtime. A few identified cases remain as false positives.

Open tasks for a follow-up PR are:

- Clarifying the status of remaining false positives
- Actually removing the commented out dependencies once it is clear that they are not needed in production.

### DIFF
--- a/ors-api/pom.xml
+++ b/ors-api/pom.xml
@@ -249,23 +249,6 @@
         </dependency>
         -->
 
-        <!-- TODO: remove this if nothing breaks -->
-        <?ignore NOTE: this non-existing processing instruction is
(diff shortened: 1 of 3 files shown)


MichaelsJP (approved):
I adjusted a couple of packages and removed some that were already flagged as removable. Everything else looks good to me!

PR #1703, opened 2024-02-28
docs: add FAQ for barrier and CH

koebi:
These two issues come up every now and then.
As I see it, there is currently no really good place to document these things.
The FAQ will have to do for now :)


aoles (approved):
Thanks a lot for taking care of this! ❤️

takb (approved):
LGTM, thanks!

PR #1708, opened 2024-03-04
fix(isochrones): correct and unify edge splitting

aoles:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

- Key functionality added: Split edges based on difference in coordinates rather than their actual distance in meters, and use the same splitting length across isochrone builders.
- Reason for change: Distance-based splitting proved to be unreliable, while different defaults across the builders introduced additional differences in the generated geometries.

In order to guarantee correct construction of the concave hull edge splitting operations need to be performed based on the differences computed on geocoordinates rather than on the actual distances in meters. In practice, this switch makes  little difference to the produced geometries and has virtually no measurable impact on performance. Typical differences in geometries observed are as follows, where blue is the geometry produced with  the previous approach, and red with the new one.

image

The reduction of the splitting length has more noticeable effects, see [report](https://rpubs.com/aoles/splitting_length_reduction). For regular isochrones going down by 1/3 from `0.012` to `0.009`, which corresponds to 1333 and 999 meters in latitude, respectively, comes with a performance penalty of around 15%. However, it helps to increase isochrone accuracy and avoid false inclusion of unreachable areas, which is best illustrated by the following example.

image
<sup>Location 10</sup>

Typically, the observed differences are more subtle, such as

image
<sup>Location 4</sup>

and

image
<sup>Location 7</sup>

### DIFF
--- a/ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java
+++ b/ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java
@@ -22,7 +22,7 @@ import java.util.List;
 import static org.locationtech.jts.algorithm.hull.ConcaveHull.concaveHullByLength;
 
 public abstract class AbstractIsochroneMapBuilder implements IsochroneMapBuilder {
-    private static final double MAX_SEGMENT_LENGTH = 0.18;// corresponds to 20000 m
+    private static final double MAX_SEGMENT_ANGLE = 0.18;// in degrees, corresponds to 20000 m
     protected static final DistanceCalc dcFast = new DistancePlaneProjection();
     protected GeometryFactory geometryFactory;
     protected RouteSearchContext searchContext;
@@ -170,7 +170,7 @@ public abstract class AbstractIsochroneMapBuilder implements IsochroneMapBuilder
                 lat1 = pl.getLat(i);
                 lon1 = pl.getLon(i);
                 addPoint(points, lon0, lat0);
-                splitLineSegment(lat0, lon0, lat1, lon1, points, minSplitLength, MAX_SEGMENT_LENGTH);
+                splitLineSegment(lat0, lon0, lat1, lon1, points, minSplitLength, MAX_SEGMENT_ANGLE);
                 lon0 = lon1;
                 lat0 = lat1;
             }
--- a/ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java
+++ b/ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java
@@ -22,7 +22,7 @@ import java.util.List;
 import static org.locationtech.jts.algorithm.hull.ConcaveHull.concaveHullByLength;
 
 public abstract class AbstractIsochroneMapBuilder implements IsochroneMapBuilder {
-    private static final double MAX_SPLIT_LENGTH = 20000.0;
+    private static final double MAX_SPLIT_LENGTH = 0.18;// corresponds to 20000 m
     protected static final DistanceCalc dcFast = new DistancePlaneProjection();
     protected GeometryFactory geometryFactory;
     protected RouteSearchContext searchContext;
@@ -141,12 +141,16 @@ public abstract class AbstractIsochroneMapBuilder implements IsochroneMapBuilder
      * @param maxlim limit above which the edge will NOT be split anymore (in meters)
      */
     protected void splitLineSegment(double lat0, double lon0, double lat1, double lon1, List<Coordinate> points, double minlim, double maxlim) {
-        double dist = dcFast.calcDist(lat0, lon0, lat1, lon1);
+        double dx = (lon1 - lon0);
+        double dy = (lat1 - lat0);
+        double normLength = Math.sqrt((dx * dx) + (dy * dy));
 
-        if (dist > minlim && dist < maxlim) {
-            int n = (int) Math.ceil(dist / minlim);
+        if (minlim < normLength && normLength < maxlim) {
+            int n = (int) Math.ceil(normLength / minlim);
+            double dx2 = dx / n;
+            double dy2 = dy / n;
             for (int i = 1; i < n; i++) {
-                addPoint(points, lon0 + i * (lon1 - lon0) / n, lat0 + i * (lat1 - lat0) / n);
+                addPoint(points, lon0 + i * dx2, lat0 + i * dy2);
             }
         }
     }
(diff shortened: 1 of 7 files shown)


sfendrich:
Could you please elaborate on how this change affects isolines across the globe? Using geo-coordinates means measuring in degrees. The distance between coordinates varies a lot depending on the location and the used coordinate system. Imagine you would invent a coordinate system having its north pole in Berlin. would the isolines still have the same shape?

ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java
```diff
@@ -22,11 +22,11 @@
 import static org.locationtech.jts.algorithm.hull.ConcaveHull.concaveHullByLength;
 
 public abstract class AbstractIsochroneMapBuilder implements IsochroneMapBuilder {
-    private static final double MAX_SPLIT_LENGTH = 20000.0;
+    private static final double MAX_SEGMENT_LENGTH = 0.18;// corresponds to 20000 m
```
sfendrich: What is the measurement unit?

ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java
```diff
@@ -141,12 +141,16 @@ protected void addPointsFromEdge(List<Coordinate> points, double isolineCost, fl
      * @param maxlim limit above which the edge will NOT be split anymore (in meters)
      */
     protected void splitLineSegment(double lat0, double lon0, double lat1, double lon1, List<Coordinate> points, double minlim, double maxlim) {
-        double dist = dcFast.calcDist(lat0, lon0, lat1, lon1);
+        double dx = (lon1 - lon0);
+        double dy = (lat1 - lat0);
+        double normLength = Math.sqrt((dx * dx) + (dy * dy));
```
sfendrich, line 146: Why can we use cartesian norm in a non-cartesian coordinate system?

aoles:
Thanks @sfendrich, that's a fair point! What this PR tries to achieve is to perform edge splitting in the same coordinate system in which the concave hull is also being computed. The concave hull algorithm is parameterized by the target maximum edge length. Edge splitting makes sure that edge segments which are input to the concave hull algorithm do not exceed this limit.  This PR does not make any change to the concave hull part which actually computes the isochrone polygons. This part has alwas been done in the geocoordinate domain. Having said this, I agree that the generated shape is influenced by the location.

sfendrich:
So, if I build the same town two times, once in Alaska and once in Cameroon, the isochrones would differ in their shape? This sounds undesirable.

aoles:
While the above is true, this is also how it has always been like. Let me stress this again: this PR does not change how the isochrones are build in general. It just makes adjustments to the logic how the edges are split, and the splitting is there only to facilitate correct results from computing the concave hull.

Having said this, I don't think this discussion should be continued here. If you are not satisfied with the current solution of generating isochrone geometries you are welcome to open an issue. Cheers!

rtroilo:
> While the above is true, this is also how it has always been like. Let me stress this again: this PR does not change how the isochrones are build in general. It just makes adjustments to the logic how the edges are split, and the splitting is there only to facilitate correct results from computing the concave hull.
> 
> Having said this, I don't think this discussion should be continued here. If you are not satisfied with the current solution of generating isochrone geometries you are welcome to open an issue. Cheers!

But I think this could have a very easy fix, just make the smoothing-factor depending an the latitude of the center. What do you think?

aoles:
@rtroilo thanks for chiming in. I've thought about this too, however, the further apart you move from the equator the higher  the discrepancy between the length in meters of one latitudinal and one longitudinal degrees. The algorithm which computes the concave hull is parametrized only with one length and does not differentiate between XY directions. So effectively you still have the same issue, or do I miss something? 🤔

sfendrich (changes requested):
Just two minor comments.

sfendrich:
So, to summarize the discussion:  

**The issue solved by this PR** is the following: If a long graph edge is used in an isochrone it should be contained in the concave hull. However, due to the length of the edge, the concave hull algorithm might inappropriately introduce a cave such that the edge is outside the hull. Therefore, we introduce additional nodes on this long edge to split it into multiple shorter ones, yielding more points to be fed into the concave hull algorithm. These additional points prevent the creation of the cave.

**Independent of this PR** it seems there is something fundamentally wrong with isochrones being computed from geo-coordinates. At the scale of a single isochrone, this is probably not relevant. But when comparing isochrones at different latitudes, the length differences of a longitude degree might affect the shape and size of the isochrone's geometry.
Here is a new issue to continue the discussion on this second topic: https://github.com/GIScience/openrouteservice/issues/1711.

ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java
aoles: Degrees.

ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java
aoles: This is based on the fact that the rest of polygon processing, in particular the concave hull algorithm, assumes cartesian coordinates. I believe for smaller distances such approximation is legitimate. In fact, this is not the only place such norm is being used. Similar computations are being performed when adding buffer points,

https://github.com/GIScience/openrouteservice/blob/a890ad2230e8594b0000a35e687683641ef5c5ba/ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java#L78-L82

 and in the contour preprocessing part.

https://github.com/GIScience/openrouteservice/blob/a890ad2230e8594b0000a35e687683641ef5c5ba/ors-engine/src/main/java/org/heigit/ors/fastisochrones/Contour.java#L385-L388

ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java
sfendrich: Then I suggest renaming it to `MAX_SEGMENT_ANGLE`.

sfendrich (approved):

ors-engine/src/main/java/org/heigit/ors/isochrones/builders/AbstractIsochroneMapBuilder.java
sfendrich: I've renamed it.

aoles: Thank you for going forward and let me apologize for not responding to you earlier, I was busy with other stuff :innocent:

MichaelsJP:
@aoles Can we merge this?

PR #1750, opened 2024-03-14
feat(docs): custom 404 page

koebi:
Since our documentation structure will change, many old links won't work anymore.
Customizing the 404 page will hopefully guide people in their confusion.


MichaelsJP (approved):
LGTM

PR #1784, opened 2024-04-21
docs: link to instructions for building from source

waldyrious:
Just a small fix to avoid a dead link. The previous URL [resulted in a 404 error](https://GIScience.github.io/openrouteservice/run-instance/installation/building-from-source).


aoles (approved):
LGTM, thanks for the fix! ❤️

PR #1807, opened 2024-06-14
feat: move profile hash creation outside GH

takb:
- Reason for change:
Move hash creation out of the ORSGraphHopper class to the RoutingProfile class, clean up more by only adding parameters that have graph changing effect


jhaeu:
This change is obsolete since we decided to use editorially named graphs in the repo instead profile hashes.

PR #1828, opened 2024-07-30
fix: docs-docker-example

MichaelsJP:
There was a " missing and the comment made the example not work in some terminals.


aoles (approved):
Awesome, thanks for the fix! 🎉

PR #1837, opened 2024-08-08
feat: improve export functionality

takb:
- trigger export by env var or runtime property
- set filename for export from variables
- refactored ORSInitContextListener and added a test


jhaeu (approved):
lgtm

PR #1875, opened 2024-10-28
fix: matrix limit not respected

TheGreatRefrigerator:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

- Key functionality added: fixed matrix 'maximum_route' limit for explicit 'all' in destiantions and/or sources
- Reason for change: matrix requests could circumvent the route limit as 'all' was counted as single entry during the check

-

-

### DIFF
--- a/ors-api/src/main/java/org/heigit/ors/api/services/MatrixService.java
+++ b/ors-api/src/main/java/org/heigit/ors/api/services/MatrixService.java
@@ -23,6 +23,7 @@ import org.springframework.stereotype.Service;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import static org.heigit.ors.api.requests.matrix.MatrixRequest.isFlexibleMode;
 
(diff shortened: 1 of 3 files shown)


ors-api/src/main/java/org/heigit/ors/api/services/MatrixService.java
```diff
@@ -55,8 +56,10 @@ public org.heigit.ors.matrix.MatrixRequest convertMatrixRequest(MatrixRequest ma
                 endpointsProperties.getMatrix().getMaximumVisitedNodes(),
                 endpointsProperties.getMatrix().getUTurnCost());
 
-        int numberOfSources = matrixRequest.getSources() == null ? matrixRequest.getLocations().size() : matrixRequest.getSources().length;
-        int numberODestinations = matrixRequest.getDestinations() == null ? matrixRequest.getLocations().size() : matrixRequest.getDestinations().length;
+        String[] sources = matrixRequest.getSources();
```
TheGreatRefrigerator, line 59: is it better to inline this?

aoles:
Is it actually desired to allow passing a String value of `"all"` as a valid value to the `sources`/`destination` parameter? To be honest, I didn't even know this was possible. I always understood these parameters as integer arrays, which if not provided defaulted to all locations, i.e. arrays of [0, n-1] where n is the number of provided `locations`. I'm not sure whether mixing different types such as strings and integers within one parameter is a good idea. So maybe the actual issue here is about allowing passing "all" as a valid parameter value, which to me sounds a bit obsolete considering that the default is to take all locations anyway.

TheGreatRefrigerator:
> Is it actually desired to allow passing a String value of `"all"` as a valid value to the `sources`/`destination` parameter? To be honest, I didn't even know this was possible. I always understood these parameters as integer arrays, which if not provided defaulted to all locations, i.e. arrays of [0, n-1] where n is the number of provided `locations`. I'm not sure whether mixing different types such as strings and integers within one parameter is a good idea. So maybe the actual issue here is about allowing passing "all" as a valid parameter value, which to me sounds a bit obsolete considering that the default is to take all locations anyway.

Can we discuss this in a separate issue please. Currently you can circumvent the config limit. That's fixed with the PR. Being able to set `all` has been this way at least since 2018. The whole array is read in as strings and converted afterwards...
Of course that's not conceptually ideal, but i don't think that's relevant for this PR.

ors-api/src/main/java/org/heigit/ors/api/services/MatrixService.java
aoles: I like the current version better 👍

aoles (approved):
LGTM, cheers! 🌈

PR #1889, opened 2024-11-08
feat: config migration and graph repo client 

MichaelsJP:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes # .

- Key functionality added:
- Reason for change:

-

-


sfendrich:
The commit message title of commit https://github.com/GIScience/openrouteservice/pull/1889/commits/6035301b24b0c0c9216dcfda319c55b2ecb723b9 is irritating. A refactoring is not supposed to be a breaking change.

TheGreatRefrigerator:
> The commit message title of commit [6035301](https://github.com/GIScience/openrouteservice/commit/6035301b24b0c0c9216dcfda319c55b2ecb723b9) is irritating. A refactoring is not supposed to be a breaking change.

The ! indicates a breaking change & the body describes the thing that is breaking:
```
BREAKING CHANGE: Structural changes in the configuration make this change incompatible with earlier versions and configurations.
```

according to the specification we are currently using, that's how a breaking change should be committed and it can be on any commit type:
https://www.conventionalcommits.org/en/v1.0.0/#examples

Or do you think the actual title confusing?
`refactor the config classes into a separate structure into ors-engine`
Would you suggest a different one?

sfendrich:
No, I think that in general "refactor" and "BREAKING CHANGE" are incompatible by their definition. I wonder whether there are rare exceptions to this?
As the commit changes the behavior it is either a fix or a feat.

TheGreatRefrigerator:
ah due to refactoring being defined by "not altering existing functionality" 😓. Yeah, agreed.
How about:
`feat(config)!: Move the config classes into a separate structure into ors-engine`

ors-config.env
```diff
@@ -40,7 +40,7 @@
 #ors.endpoints.matrix.maximum_routes_flexible=25
 #ors.endpoints.matrix.maximum_visited_nodes=100000
 #ors.endpoints.matrix.maximum_search_radius=2000
-#ors.endpoints.matrix.u_turn_costs=-1
+#ors.endpoints.matrix.u_turn_cost=-1
```
sfendrich: Why has this changed? We use the plural form of "costs" everywhere else.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/GreenIndexGraphStorageBuilder.java
```diff
@@ -45,10 +48,18 @@ public GraphExtension init(GraphHopper graphhopper) throws Exception {
         if (storage != null)
             throw new Exception("GraphStorageBuilder has been already initialized.");
 
-        // TODO Refactoring Check if the _greenIndexFile exists
-        String csvFile = parameters.get("filepath");
-        readGreenIndicesFromCSV(csvFile);
-        prepareGreenIndexSlots();
+        ExtendedStorageProperties parameters;
+        try {
+            parameters = this.parameters;
```
sfendrich: How is this assignment supposed to fail? The same question applies to the other GraphStoragBuilders.

ors-api/src/test/files/heidelberg.osm.gz
sfendrich, line 1: Why has the format been changed? I think XML was on purpose as the file is manually adapted to account for special test cases. XML would make it easier to inspect changes or add new changes for new tests.

ors-api/src/main/java/org/heigit/ors/api/config/EngineProperties.java
```diff
@@ -0,0 +1,9 @@
+package org.heigit.ors.api.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConfigurationProperties(prefix = "ors.engine")
+public class EngineProperties extends org.heigit.ors.config.EngineProperties {
```
sfendrich: I don't like that the child class has the same name as the parent class.

ors-engine/src/main/java/org/heigit/ors/config/ElevationProperties.java
```diff
@@ -0,0 +1,27 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import org.heigit.ors.common.DataAccessEnum;
+
+import java.nio.file.Path;
+
+@Getter
+@Setter
+public class ElevationProperties {
+    private Boolean preprocessed;
```
sfendrich, line 12: Can this be a `bool`? `Boolean`might be null...

ors-engine/src/main/java/org/heigit/ors/config/ElevationProperties.java
```diff
@@ -0,0 +1,27 @@
...
+    private DataAccessEnum dataAccess;
+    private Boolean cacheClear;
```
sfendrich, line 14: Can this be a `bool`? `Boolean`might be null...

ors-engine/src/main/java/org/heigit/ors/config/EngineProperties.java
```diff
@@ -0,0 +1,37 @@
...
+import org.heigit.ors.config.profile.ProfileProperties;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+@Getter
+@Setter
+public class EngineProperties {
+    private Integer initThreads;
+    private Boolean preparationMode;
```
sfendrich, line 16: Can this be a `bool`? `Boolean`might be null...

ors-api/pom.xml
```diff
@@ -116,13 +116,15 @@
...
 
-        <!-- TODO: This dependency is shown as used by `maven dependency:analyze`
-        although it is a transitive dependency. Find out how to resolve this as
-        we would prefer to not need to declare it here. -->
         <dependency>
-            <groupId>com.graphhopper.external</groupId>
-            <artifactId>jackson-datatype-jts</artifactId>
-            <version>0.12-2.5-1</version>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
```
sfendrich, line 121: Generally, I'm skeptical about the introduction of lombok, because I think that its benefits do not outweigh the risk. I can see some benefit for some of the configuration classes, where many getters/setters might be necessary. However, we should first evaluate whether other approaches such as final fields, records, etc. can solve the same issue.
There are some places, where I particularly vote against using lombok, which I mark with further comments.

ors-api/src/main/java/org/heigit/ors/api/requests/common/APIRequest.java
```diff
@@ -19,6 +21,10 @@ public class APIRequest {
     @Schema(name = PARAM_PROFILE, hidden = true)
     protected APIEnums.Profile profile;
 
+    @Getter
+    @Setter
+    protected String profileName;
```
sfendrich, line 23: Lombok should not be used here because:

1. It reduces only a single getter/setter
2. The usage of this field is semantically not a getter/setter

ors-engine/src/main/java/org/heigit/ors/common/EncoderNameEnum.java
```diff
@@ -0,0 +1,191 @@
...
+    GH_BIKE("gh-bike", 42),
+    GH_BIKE2("gh-bike2", 43),
+    GH_BIKE_MTB("gh-mtb", 44),
+    GH_BIKE_ROAD("gh-racingbike", 45),
+    GH_FOOT("gh-foot", 46),
+    GH_HIKE("gh-hike", 47);
+
+    public final String name;
+    @Getter
+    private final int value;
```
sfendrich, line 55: Why use lombok here to remove only one getter?

ors-engine/src/main/java/org/heigit/ors/export/ExportRequest.java
```diff
@@ -33,6 +35,9 @@ public class ExportRequest extends ServiceRequest {
     private static final Logger LOGGER = Logger.getLogger(ExportRequest.class);
     private BBox boundingBox;
 
+    @Getter
+    @Setter
+    private String profileName;
```
sfendrich, line 36: Lombok should not be used here because:

1. It reduces only a single getter/setter
2. The usage of this field is semantically not a getter/setter

ors-engine/src/main/java/org/heigit/ors/isochrones/IsochroneRequest.java
```diff
@@ -24,6 +26,9 @@
 import java.util.*;
 
 public class IsochroneRequest extends ServiceRequest {
+    @Getter
+    @Setter
+    private String profileName;
```
sfendrich, line 27: Lombok should not be used here because:

1. It reduces only a single getter/setter
2. The usage of this field is semantically not a getter/setter

ors-engine/src/main/java/org/heigit/ors/matrix/MatrixRequest.java
```diff
@@ -40,6 +42,9 @@
 import org.locationtech.jts.geom.Coordinate;
 
 public class MatrixRequest extends ServiceRequest {
+    @Getter
+    @Setter
+    private String profileName;
```
sfendrich, line 43: Lombok should not be used here because:

1. It reduces only a single getter/setter
2. The usage of this field is semantically not a getter/setter

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/GraphProcessContext.java
```diff
@@ -31,55 +32,38 @@
 
 public class GraphProcessContext {
     private static final Logger LOGGER = Logger.getLogger(GraphProcessContext.class.getName());
-    private List<GraphBuilder> graphBuilders;
-    private GraphBuilder[] arrGraphBuilders;
+    @Getter
```
sfendrich: Lombok should not be used here because:

1. It reduces only a single getter/setter
2. The usage of this field is semantically not a getter/setter

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/GraphProcessContext.java
```diff
@@ -31,55 +32,38 @@
...
     private List<GraphStorageBuilder> storageBuilders;
     private GraphStorageBuilder[] arrStorageBuilders;
     private int trafficArrStorageBuilderLocation = -1;
+    @Getter
     private final double maximumSpeedLowerBound;
```
sfendrich, line 37: Lombok should not be used here because:

1. It reduces only a single getter/setter
2. The usage of this field is semantically not a getter/setter

ors-engine/src/main/java/org/heigit/ors/snapping/SnappingRequest.java
```diff
@@ -6,17 +6,22 @@
...
 import org.heigit.ors.routing.WeightingMethod;
 import org.heigit.ors.routing.graphhopper.extensions.ORSWeightingFactory;
 import org.heigit.ors.util.ProfileTools;
-import org.heigit.ors.routing.RoutingProfile;
 import org.locationtech.jts.geom.Coordinate;
 
 public class SnappingRequest extends ServiceRequest {
+    @Getter
+    @Setter
+    private String profileName;
```
sfendrich, line 20: Lombok should not be used here because:

1. It reduces only a single getter/setter
2. The usage of this field is semantically not a getter/setter

sfendrich (changes requested):
This is only a partial review suggesting some improvements. The PR is too large to conduct a full review.

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
```diff
@@ -0,0 +1,205 @@
...
+
+    public void setIsActivatingGraphs(boolean value) {
+        isActivatingGraphs.set(value);
+    }
+
+    @Async
+    @Scheduled(cron = "${ors.engine.graph_management.download_schedule:0 0 0 31 2 *}")//Default is "never"
+    public void checkForUpdatesInRepo() {
+
+        if (!enabled) {
```
sfendrich: could be null

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Async
+    @Scheduled(cron = "${ors.engine.graph_management.activation_schedule:0 0 0 31 2 *}")//Default is "never"
+    public void checkForDownloadedGraphsToActivate() {
+        checkForDownloadedGraphsToActivate("Scheduled");
+    }
+
+    public void checkForDownloadedGraphsToActivate(String trigger) {
+        if (!enabled) {
```
sfendrich: could be null

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        LOGGER.info("%s graph activation check done: Performing graph activation...".formatted(trigger));
+        activateGraphs();
+    }
+
+    @Async
+    @Scheduled(fixedDelay = 1, timeUnit = TimeUnit.MINUTES)
+    public void repeatedGraphActivationAttempt() {
+        if (!enabled) {
```
sfendrich: could be null

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Async
+    @Scheduled(cron = "${ors.engine.graph_management.download_schedule:0 0 0 31 2 *}")//Default is "never"
+    public void checkForUpdatesInRepo() {
+
+        if (!enabled) {
+            LOGGER.debug("Graph management is disabled, skipping scheduled repository check...");
+            return;
+        }
```
sfendrich, line 58: We repeatedly have to check for being enabled. Shouldn't this be handled through polymorphism?

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/manage/remote/AbstractRepoManager.java
sfendrich, line 1: What makes this class and its child classes RepoManagers instead of Repos?

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/manage/PersistedGraphInfo.java
sfendrich, line 1: GraphInfo is a very generic name. Can we provide a better one that expresses the content of this class and its subclasses?

sfendrich (changes requested):
Partial review.

ors-engine/src/main/java/org/heigit/ors/config/ElevationProperties.java
MichaelsJP: This is a config Jackson serialization class. The property needs to be nullable.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/manage/remote/FileSystemRepoManager.java
```diff
@@ -0,0 +1,133 @@
...
+        if (LOGGER.isTraceEnabled()) {
+            LOGGER.trace("[%s] Downloading %s to local file %s...".formatted(getProfileDescriptiveName(), repoPath.toFile().getAbsolutePath(), localPath.getAbsolutePath()));
+        } else {
+            LOGGER.info("[%s] Downloading %s...".formatted(getProfileDescriptiveName(), repoPath.toFile().getName()));
+        }
+        if (repoPath != null) {
+            try {
+                Files.copy(repoPath, localPath.toPath(), StandardCopyOption.REPLACE_EXISTING);
+            } catch (IOException e) {
+                LOGGER.warn("[%s] Caught %s when trying to download %s".formatted(getProfileDescriptiveName(), repoPath.toFile().getAbsolutePath()));
```
sfendrich: The format string contains three parameters %s but only two are used. Might result in undefined behavior.

ors-api/src/main/java/org/heigit/ors/api/config/EndpointsProperties.java
```diff
@@ -0,0 +1,143 @@
...
+    @Setter(AccessLevel.PACKAGE)
+    public static class EndpointMatrixProperties {
+        private boolean enabled;
+        private String attribution;
+        private int maximumRoutes;
+        private int maximumRoutesFlexible;
+        private int maximumVisitedNodes;
+        private double maximumSearchRadius;
+        // TODO: this parameter is only used in a binary check for infinity (==-1);
+        //       Can't we reduce it to a boolean "forbid_u_turns"?
```
sfendrich: This should not be the case. GH is actually able to use a finite cost-value for u-turns. If this is disabled on the ORS side, I would consider it a bug.

sfendrich (changes requested):
A few more comments

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/manage/remote/FileSystemRepoManager.java
MichaelsJP: Fixed in another revision already.

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
MichaelsJP: refactored to `Boolean.FALSE.equals(enabled)`.

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
MichaelsJP: Fixed with `Boolean.FALSE.equals(enabled)`

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
MichaelsJP: Fixed with `Boolean.FALSE.equals(enabled)`

ors-api/src/main/java/org/heigit/ors/api/config/EndpointsProperties.java
MichaelsJP: I agree on @sfendrich argument. Let's keep it like this. I'll remove the TODO.

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
MichaelsJP: How would you implement polymorphism with this class?

ors-engine/src/main/java/org/heigit/ors/snapping/SnappingRequest.java
MichaelsJP: 1. Good point. There is even more potential. All properties are accessed from the outside and have boilerplate getter at the moment. Let's add a global @Getter. A setter is only needed for profileName and maximumLocations since the other properties are set in the constructor.
2. Please elaborate on this, I don't understand.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/GraphProcessContext.java
MichaelsJP: > It reduces only a single getter/setter
Yes. We should refactor more into this class.
> The usage of this field is semantically not a getter/setter
Please elaborate on this.

ors-engine/src/main/java/org/heigit/ors/matrix/MatrixRequest.java
MichaelsJP: > It reduces only a single getter/setter
Yes. We should refactor more into this class.
> The usage of this field is semantically not a getter/setter
Please elaborate on this.

ors-engine/src/main/java/org/heigit/ors/isochrones/IsochroneRequest.java
MichaelsJP: > It reduces only a single getter/setter

Yes. We should refactor more into this class.

> The usage of this field is semantically not a getter/setter

Please elaborate on this.

ors-api/pom.xml
MichaelsJP: We deal with a lot of boilerplate code, especially for tasks like getters, setters, and constructors—trivial code that rarely gets tested and often becomes a liability. Eliminating this clutter would significantly improve code readability and maintainability, allowing us to focus on core business logic. Lombok is designed to address precisely this issue.

> I think that its benefits do not outweigh the risk.

I understand the concern, especially with introducing new libraries. However, Lombok has grown into a widely adopted and mature tool with strong community and industry support. Debating its adoption at this point feels somewhat outdated. A practical approach could be to limit its initial use to new code, focusing on basic annotations like @Getter, @Setter, @NoArgsConstructor, @AllArgsConstructor, and @Builder, while avoiding changes to legacy code during a trial phase.

> We should first evaluate whether other approaches such as final fields, records, etc., can solve the same issue.

Final fields and records can certainly help in some cases, but they don’t fully replace Lombok. Records, for example, are great for specific scenarios but come with notable limitations: no setters, no inheritance, and a fixed state. These constraints make them impractical for many use cases. In my opinion, we can evaluate this on a case-by-case basis and continue to use native Java features wherever they make sense.

ors-engine/src/main/java/org/heigit/ors/config/EngineProperties.java
MichaelsJP: It needs to be null for the whole Jackson/spring config reading functionality. Else, we cannot differentiate what the user has set and what needs to be populated with our default values.

ors-engine/src/main/java/org/heigit/ors/config/ElevationProperties.java
MichaelsJP: It needs to be null for the whole Jackson/spring config reading functionality. Else, we cannot differentiate what the user has set and what needs to be populated with our default values.

ors-config.env
MichaelsJP: Agree. This should be changed back. Probably happened by mistake.

ors-api/src/test/files/heidelberg.osm.gz
MichaelsJP: The original `.osm.gz` was a compressed binary, no? 
Tracking git changes is as possible as with osm.pbf.
The reading of .osm.pbf and therefore the creation of graphs is faster. That was mainly the reason, I guess.

ors-api/src/main/java/org/heigit/ors/api/config/EngineProperties.java
MichaelsJP: Agree. I will rename it to `ApiEngineProperties`

ors-api/pom.xml
MichaelsJP: I agree that introducing Lombok is a certain step. To that end, @jhaeu , @takb , and I made this decision together after a thorough discussion of the pros and cons. It wasn’t a spontaneous choice. @jhaeu 's years of experience with Lombok definitely helped to get another perspective on it.

Given that not everyone fully agrees on adopting Lombok, we’ve decided to limit its usage for now. For the moment we stick to a base set of annotations:  `@Getter`, `@Setter`, `@NoArgsConstructor`, `@AllArgsConstructor`, `@Builder`, and `@Accessors`. Those have already shown in the current context to really increase the practicability and readability of our code.

To address concerns, we could agree to restrict Lombok usage to new code, touching legacy code only during specific refactoring efforts—not as a blanket change. (Apologies for my inconsistency with some of the getter/setter annotations further down; Those were initial tryouts, I’ll clean those up! 😉)

That said, I still have a few questions further down regarding other comments.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/GreenIndexGraphStorageBuilder.java
MichaelsJP: It doesn't. It might be a remnant of earlier code. Will remove it.

MichaelsJP: https://github.com/GIScience/openrouteservice/pull/1908

ors-api/src/main/java/org/heigit/ors/api/config/EngineProperties.java
MichaelsJP: https://github.com/GIScience/openrouteservice/pull/1908

ors-engine/src/main/java/org/heigit/ors/isochrones/IsochroneRequest.java
MichaelsJP: https://github.com/GIScience/openrouteservice/pull/1908

ors-api/src/main/java/org/heigit/ors/api/requests/common/APIRequest.java
MichaelsJP: 1. Yes. Will remove it. Also, not PR related legacy code.

2. Not a real discussion point, but please explain why 2 is the case. I don't understand why profileName is not a semantic getter/setter while the whole class has dedicated getter/setter boilerplate on all fields?

MichaelsJP: I closed the same discussion below with the following PR that solves all of them:

https://github.com/GIScience/openrouteservice/pull/1889

ors-engine/src/main/java/org/heigit/ors/common/EncoderNameEnum.java
MichaelsJP: https://github.com/GIScience/openrouteservice/pull/1889

ors-engine/src/main/java/org/heigit/ors/export/ExportRequest.java
MichaelsJP: https://github.com/GIScience/openrouteservice/pull/1889

ors-engine/src/main/java/org/heigit/ors/matrix/MatrixRequest.java
MichaelsJP: https://github.com/GIScience/openrouteservice/pull/1889

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/GraphProcessContext.java
MichaelsJP: https://github.com/GIScience/openrouteservice/pull/1889

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/GraphProcessContext.java
MichaelsJP: https://github.com/GIScience/openrouteservice/pull/1908

ors-engine/src/main/java/org/heigit/ors/snapping/SnappingRequest.java
MichaelsJP: https://github.com/GIScience/openrouteservice/pull/1908

ors-api/src/test/files/heidelberg.osm.gz
sfendrich: It is a gzipped XML, so yes, you cannot track changes directly in git, but it easy to unpack and diff. And adding new scenarios in the future is also easier.
Heidelberg is such a small area, do we really have a noticable performance difference?

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
sfendrich: We could have two different graph-service classes. One for the enabled graph service and one no-op graph service. So there is only one check when instantiating the graph service from the config parameter.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/manage/remote/FileSystemRepoManager.java
sfendrich: Where?

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
jhaeu: Ok, I can try to do this

ors-api/src/test/files/heidelberg.osm.gz
sfendrich: In an internal discussion we decided to keep the new pbf variant for performance reasons. In case we need to add further data, we can vonvert it to XML and back again.

ors-api/src/main/java/org/heigit/ors/api/services/GraphService.java
sfendrich: As this is a purely cosmetic issue, let's postpone the discussion about this until after the release.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/manage/remote/AbstractRepoManager.java
MichaelsJP: Valid point. The `*RepoManger` classes are now renamed to `*GraphRepoClient` classes. Makes it more precise.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/manage/PersistedGraphInfo.java
MichaelsJP: Good point. It is now renamed to `GraphInfo` -> `GraphBuildInfo`. This highlights its actual usage as being part of the graph build process and providing more information around it as part of the new `graph_build_info.yml` in the graph folders.

sfendrich (approved):

jhaeu (approved):
LGTM

PR #1954, opened 2025-01-22
fix: Implicit vehicle access handling for ferries

koebi:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Not sure if there are any cases that would be covered by the `or`-part of the original if-statement that are *not* caught by the new statement…

One would probably have to change `hgv` as well, I think?

- Key functionality added: Use ferries that do not have any general or vehicle-specific access tag (such as `foot`, `motorcar`, `access`, …), but only the `route=ferry` tag.
- Reason for change: According to the [tag filtering logic](https://giscience.github.io/openrouteservice/technical-details/tag-filtering) we accept these ways, which I think makes sense.
Whether it makes sense to only do that if `foot` or `bike` are absent, I don't know…

[This route](https://classic-maps.openrouteservice.org/directions?n1=58.238978&n2=6.621854&n3=17&a=58.241759,6.631633,58.236431,6.612058&b=0&c=0&k1=en-US&k2=km) using [such a ferry](https://www.openstreetmap.org/way/28251389/) in norway will now be available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ RELEASING:
 - NPE in error handling ([#1925](https://github.com/GIScience/openrouteservice/pull/1925))
 - Add missing 'build' in documentation for profile
   properties ([#1947](https://github.com/GIScience/openrouteservice/issues//1947))
-- implicit vehicle access handling for ferries
+- Allow vehicles on ferries without explicit access tags ([#1954](https://github.com/GIScience/openrouteservice/pull/1954))
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ RELEASING:
 - NPE in error handling ([#1925](https://github.com/GIScience/openrouteservice/pull/1925))
 - Add missing 'build' in documentation for profile
   properties ([#1947](https://github.com/GIScience/openrouteservice/issues//1947))
+- implicit vehicle access handling for ferries
 
 ### Security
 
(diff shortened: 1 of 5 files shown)


aoles:
FYI: The logic to allow access only in the absence of `foot` or `bike` tags has been adopted from graphhopper and has been originally introduced in https://github.com/graphhopper/graphhopper/commit/f87fba46294ed56a6f24a7374b7e34290be3d6e8.

aoles (changes requested):
Many thanks for the prompt fix! 🚀 

Looks aready quite good. Would you mind adding:
- corresponding changes to the HGV flag encoder/tests
- an entry in the CHANGELOG file?

Cheers! ❤️

aoles:
Hi @koebi,
thanks for implementing the suggested changes! Unfortunately, a few API tests started failing, some with quite big deviations from the expected values 💀  I'm a bit puzzled by this, especially as this PR effectively restores the original behavior that got broken in https://github.com/GIScience/openrouteservice/pull/1291. There the influence on API tests was negligible.

aoles:
Some of the proposed changes seem to break the routing 💀  This is evident e.g. when looking at the route from `testHGVHazmatRestriction` api test for `hazmat = true` (blue is the reference as on main, and red the results calculated with this branch)

image

Apparently some sort of secret tram routing option got activated, see below 😂 

image

Not sure what the following is about though 🙈 

image

aoles (approved):
LGTM! 🚀

PR #1970, opened 2025-02-17
fix: correct matrix error output

takb:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

- Reason for change:
Some error responses of the matrix endpoint were not formed correctly using the internal error codes.

### DIFF
--- a/ors-engine/src/main/java/org/heigit/ors/matrix/MatrixRequest.java
+++ b/ors-engine/src/main/java/org/heigit/ors/matrix/MatrixRequest.java
@@ -23,9 +23,7 @@ import com.graphhopper.storage.RoutingCHGraph;
 import com.graphhopper.util.PMap;
 import org.heigit.ors.common.DistanceUnit;
 import org.heigit.ors.common.ServiceRequest;
-import org.heigit.ors.exceptions.InternalServerException;
-import org.heigit.ors.exceptions.MaxVisitedNodesExceededException;
-import org.heigit.ors.exceptions.PointNotFoundException;
+import org.heigit.ors.exceptions.*;
 import org.heigit.ors.matrix.algorithms.core.CoreMatrixAlgorithm;
 import org.heigit.ors.matrix.algorithms.dijkstra.DijkstraMatrixAlgorithm;
 import org.heigit.ors.matrix.algorithms.rphast.RPHASTMatrixAlgorithm;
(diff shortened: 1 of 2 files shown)


ors-engine/src/main/java/org/heigit/ors/matrix/MatrixRequest.java
```diff
@@ -224,9 +222,9 @@ else if (getSearchParameters().getDynamicSpeeds() && routingProfile.getGraphhopp
         } catch (PointNotFoundException e) {
             throw e;
         } catch (MaxVisitedNodesExceededException e) {
-            throw new InternalServerException(MatrixErrorCodes.MAX_VISITED_NODES_EXCEEDED, "Unable to compute a distance/duration matrix: " + e.getMessage());
+            throw new StatusCodeException(400, MatrixErrorCodes.MAX_VISITED_NODES_EXCEEDED, "Unable to compute a distance/duration matrix: " + e.getMessage());
```
aoles: According to [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) 4xx status codes indicate client errors. However, exceeding of the maximum allowed visited nodes is a server-side error. As such, it should return a 5xx error.

aoles (changes requested):
Please see my inline comment. Cheers! 🤟

ors-engine/src/main/java/org/heigit/ors/matrix/MatrixRequest.java
takb: fixed.

ors-engine/src/main/java/org/heigit/ors/matrix/MatrixRequest.java
```diff
@@ -224,9 +222,9 @@ else if (getSearchParameters().getDynamicSpeeds() && routingProfile.getGraphhopp
         } catch (PointNotFoundException e) {
             throw e;
         } catch (MaxVisitedNodesExceededException e) {
-            throw new InternalServerException(MatrixErrorCodes.MAX_VISITED_NODES_EXCEEDED, "Unable to compute a distance/duration matrix: " + e.getMessage());
+            throw new StatusCodeException(500, MatrixErrorCodes.MAX_VISITED_NODES_EXCEEDED, "Unable to compute a distance/duration matrix: " + e.getMessage());
```
aoles, line 225: Am I missing something here, or is this just a refactoring? According to 

https://github.com/GIScience/openrouteservice/blob/6cf11dbc2cc12fdc1191f2eab06c50f07cd18840/ors-engine/src/main/java/org/heigit/ors/exceptions/InternalServerException.java#L29-L31

a call to `StatusCodeException(500, MatrixErrorCodes.MAX_VISITED_NODES_EXCEEDED, "...")` is effectively the same as `InternalServerException(MatrixErrorCodes.MAX_VISITED_NODES_EXCEEDED, "...")`.

PR #2047, opened 2025-04-25
feat: add access = permit to known access restiction types

### LINKED ISSUE #2013: Add `access = permit` to known access restiction types
### Feature/Enhancement Proposal

Information about restricted usage for roads with access restriction of type [permit](https://wiki.openstreetmap.org/wiki/Tag:access%3Dpermit) is not reported because the restriction is missing from the list of known [restriction types](https://github.com/GIScience/openrouteservice/blob/f401ccf0a345384f310b8c39305d19c05243fc85/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/AccessRestrictionType.java).

As an example, consider [Eternitstraße](https://www.openstreetmap.org/way/176768688) in Heidelberg.

Image

Google Maps correctly reports it as restricted usage road.

Image

merydian:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.
- [x] 7. If changes/additions are made to the ors-config.json file, I have added these to the [ors config documentation][config]
- [x] 13. For changes touching the API documentation, I have tested that the API playground [renders correctly][api].

Fixes #2013.

- Key functionality added: Access restrictions with the type "permit" will now be considered during graph building. 
- Reason for change:

-

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Releasing is documented in RELEASE.md
 
 ### Added
 - add hint to use yaml editor in docs ([#1836](https://github.com/GIScience/openrouteservice/issues/1836))
+- access = permit as known access restriction type ([#2013](https://github.com/GIScience/openrouteservice/issues/2013))
 
 ### Changed
 
--- a/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilder.java
+++ b/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilder.java
@@ -78,9 +78,9 @@ public class RoadAccessRestrictionsGraphStorageBuilder extends AbstractGraphStor
      * @return The RoadAccessRestrictionStorage object created as part of the initialisation
      * @throws Exception
      */
-    public GraphExtension init(GraphHopper graphhopper, int profileType) throws Exception {
+    public GraphExtension init(GraphHopper graphhopper, int profileType) throws IllegalStateException {
         if (storage != null)
-            throw new Exception("GraphStorageBuilder has been already initialized.");
+            throw new IllegalStateException("GraphStorageBuilder has been already initialized.");
 
         ExtendedStorageProperties parameters;
         parameters = this.parameters;
@@ -102,9 +102,9 @@ public class RoadAccessRestrictionsGraphStorageBuilder extends AbstractGraphStor
      * @return The RoadAccessRestrictionStorage object created as part of the initialisation
      * @throws Exception
      */
-    public GraphExtension init(GraphHopper graphhopper) throws Exception {
+    public GraphExtension init(GraphHopper graphhopper) throws IllegalStateException {
         if (storage != null)
-            throw new Exception("GraphStorageBuilder has been already initialized.");
+            throw new IllegalStateException("GraphStorageBuilder has been already initialized.");
 
         ExtendedStorageProperties parameters;
         parameters = this.parameters;
--- a/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilder.java
+++ b/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilder.java
@@ -61,6 +61,7 @@ public class RoadAccessRestrictionsGraphStorageBuilder extends AbstractGraphStor
         restrictedValues.add("customers");
         restrictedValues.add("emergency");
         restrictedValues.add("permissive");
+        restrictedValues.add("delivery");
         restrictedValues.add("permit");
 
         permissiveValues.add("yes");
--- a/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilder.java
+++ b/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilder.java
@@ -61,6 +61,7 @@ public class RoadAccessRestrictionsGraphStorageBuilder extends AbstractGraphStor
         restrictedValues.add("customers");
         restrictedValues.add("emergency");
         restrictedValues.add("permissive");
+        restrictedValues.add("permit");
 
         permissiveValues.add("yes");
         permissiveValues.add("designated");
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ Releasing is documented in RELEASE.md
 
 ### Added
 - add hint to use yaml editor in docs ([#1836](https://github.com/GIScience/openrouteservice/issues/1836))
+- Add access = permit to known access restiction types ([#2013](https://github.com/GIScience/openrouteservice/issues/2013))
```
aoles: Please refrain from starting the entries under *Added* with "add" (similar applies to the category *Fixed* where the entries should avoid repeating repeat "fix"). It's clear from the context that something has been added (or fixed) and the actually interesting part is **what**. Also, please consider formatting tag values as code snippets like `access = permit`. Let's keep the CHANGELOG clean and concise! 💪

ors-engine/src/test/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilderTest.java
```diff
@@ -47,6 +47,11 @@ void testCarWayCreation() throws Exception {
         way.setTag("motorcar", "yes");
         builder.processWay(way);
         assertEquals(AccessRestrictionType.NONE, builder.getRestrictions());
+
+        way = new ReaderWay(1);
+        way.setTag("access", "permit");
+        builder.processWay(way);
+        assertEquals(AccessRestrictionType.PERMIT, builder.getRestrictions());
```
aoles: I've added a commit that introduces a test intended to cover all values of `AccessRestrictionType` for an unspecified vehicle, so let's leave the original car-specific test `testCarWayCreation` untouched.

The newly introduced test fails, which is expected in the current implementation. To make it work, please adapt the code of `RoadAccessRestrictionsGraphStorageBuilder` to account for `RoutingProfileType.UNKNOWN`. In this process, you might discover some hidden bugs in the current implementation... Have fun! 😉 

https://github.com/GIScience/openrouteservice/blob/78c0612c7d1c8de36185a875e222d1eced1a80b1/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilder.java#L160-L170

aoles (changes requested):
Thanks a lot for you prompt PR ❤️ 

Please have a look at my comments, and address the issues reported by Quality Gate.

Cheers!

ors-engine/src/main/java/org/heigit/ors/routing/RoutingProfileType.java
```diff
@@ -104,6 +104,10 @@ public static boolean isCycling(int routePref) {
                 || routePref == GH_BIKE_ROAD;
     }
 
+    public static boolean isUNKNOWN(int routePref) {
```
aoles: I would personally refrain from introducing this type of trivial predicate method, especially if it's only used once.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilder.java
```diff
@@ -166,6 +168,8 @@ public void processWay(ReaderWay way, Coordinate[] coords, Map<Integer, Map<Stri
                 restrictions = isAccessAllowed(way, VAL_BICYCLE) ? 0 : getRestrictionType(way, VAL_BICYCLE);
             if (RoutingProfileType.isPedestrian(profileType))
                 restrictions = isAccessAllowed(way, "foot") ? 0 : getRestrictionType(way, "foot");
+            if (RoutingProfileType.isUNKNOWN(profileType))
```
aoles: Please use a direct check against `profileType == RoutingProfileType.isUNKNOWN` rather than introducing a new method.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilder.java
```diff
@@ -166,6 +168,8 @@ public void processWay(ReaderWay way, Coordinate[] coords, Map<Integer, Map<Stri
...
+                restrictions = isAccessAllowed(way, "unknown") ? 0 : getRestrictionType(way, "unknown");
```
aoles: The second argument to `isAccessAllowed` and `getRestrictionType` corresponds to OSM access-type key, as described in the [OSM wiki](https://wiki.openstreetmap.org/wiki/Key:access). In its most general form it specifies `access` for all modes of transport, e.g. `access = no` says that nobody may legally access the road. This can be overridden by additional tags defining mode of transport such as `bike = yes`. In this context "unknown" is not a valid value.

What I envisioned was to have a fallback value which does not depend on a specific mode of transport, i.e. which just checks based on the value of the `access` tag. For this, it should be enough to have
```java
restrictions = getRestrictionType(way, "access");
```

aoles (changes requested):
Many thanks for addressing my comments! 👍 

Please have a look at some additional remarks in the comments below.

The maintainability [issue](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=2047&id=GIScience_openrouteservice&open=AZZtCKvLTKPQaCOl0Qm-) reported by SonarCloud should be possible to address by changing the general `Exception` thrown to a more specific one such as `IllegalStateException`.

Cheers! ❤️

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/RoadAccessRestrictionsGraphStorageBuilder.java
```diff
@@ -166,6 +168,8 @@ public void processWay(ReaderWay way, Coordinate[] coords, Map<Integer, Map<Stri
                 restrictions = isAccessAllowed(way, VAL_BICYCLE) ? 0 : getRestrictionType(way, VAL_BICYCLE);
             if (RoutingProfileType.isPedestrian(profileType))
                 restrictions = isAccessAllowed(way, "foot") ? 0 : getRestrictionType(way, "foot");
+            if (profileType == RoutingProfileType.UNKNOWN)
+                restrictions = getRestrictionType(way, "access");;
```
aoles: You might want to fix the double `;;` 😅

aoles (approved):
Thanks for bearing with me! ❤️

PR #2060, opened 2025-05-16
fix: avoid roads reserved to customers

### LINKED ISSUE #1981: Directions API routes through parking lots rather than go through intersections

- [x] I have searched the existing issues

### Problem description

For the record I first posted about this issue on the forum ([link](https://ask.openrouteservice.org/t/problematic-routing-from-ors/7089)) but I figured it made more sense to create an issue here. This issue will more or less be a copy paste of that forum post.
<br>

Basically, when using the Directions API (self-hosted through Docker), sometimes the route returned by ORS takes ‘shortcuts’ through parking lots (and other similar type "roads"), instead of going through the proper intersections.
This is an issue since, unless I am mistaken, where I live this is illegal to do. Its also typically much slower due to the much lower speed limit in parking lots + the presence of pedestrians. Also, it results in an overall higher number of turns compared to using the ‘normal’ intersection…

I first saw this issue on a self hosted (Docker) ORS instance. I tried on 3 computers (1 Linux, two Windows, all through Docker so the host OS should not have any impact regarding this issue), with the latest OSM data for the region (Canada/Québec).

I then tried and got the same result with ORS’s online map service. Here are some simple examples of what can happen:
[Example 1](https://classic-maps.openrouteservice.org/directions?n1=45.645151&n2=-73.860172&n3=18&a=45.645818,-73.858595,45.644663,-73.862629&b=0&c=0&k1=en-US&k2=km)
[Example 2](https://classic-maps.openrouteservice.org/directions?n1=45.644719&n2=-73.859426&n3=18&a=45.645818,-73.858595,45.644341,-73.86154&b=0&c=0&k1=en-US&k2=km)
[Example 3](https://classic-maps.openrouteservice.org/directions?n1=45.645203&n2=-73.868294&n3=18&a=45.645796,-73.869088,45.644693,-73.869361&b=0&c=0&k1=en-US&k2=km)
[Example 4](https://classic-maps.openrouteservice.org/directions?n1=45.643898&n2=-73.869045&n3=18&a=45.644802,-73.868964,45.643399,-73.870059&b=0&c=0&k1=en-US&k2=km)

In all those cases, the proposed route is worst then simply following the ‘real’ road due to traversing parking lots.

Now, for the record, I tried those very same examples on OpenStreetMap’s website using all three available routing engines (Valhalla, GraphHopper, OSRM) and they all gave a ‘normal’ result that uses the intersection, as expected.

<br>

My best guess as to why this is happening is that those roads are marked with the OSM tag “highway=service” (or at least the ones I checked) but without the “service=parking_aisle". Some of the ones I tested had “service=driveway”. From what I can tell, they all had “access=customer”.

From my understanding of the OSM tags documentation, all of those should have been avoided, unless absolutely needed (like if the start/end point was only accessible through such road).

<br>

I also tried to find some settings in the API or within the different driving profiles, but I haven’t found anything that allowed me to avoid such results.

Am I doing something wrong? Did I truly find a bug, and if so, where should I start looking to try and fix this (if I figure out how to that is)?
I haven't really coded in Java in a while but i might be able to look into it a little.

<br>
<br>

PS: Sorry, this is so long, I tried to give all the details I could think of… But i figured its probably better to give more than less.

<br>
<br>

Also, just to be clear, it does not do that on every corner with parking lots and/or businesses/shops. I did find some that do work properly like this one here:
[Properly working corner example](https://classic-maps.openrouteservice.org/directions?n1=45.643771&n2=-73.858868&n3=18&a=45.643336,-73.858461,45.644521,-73.859217&b=0&c=0&k1=en-US&k2=km)

And here is a somewhat ironic case where it tries very very hard to avoid going through the parking lot, even though it starts from within that very same parking lot:
[Peculiar routing](https://classic-maps.openrouteservice.org/directions?n1=45.639896&n2=-73.842764&n3=18&a=45.640016,-73.842491,45.640316,-73.843516&b=0&c=0&k1=en-US&k2=km)
I still included it as it might be linked to this issue somehow?

Also, there was this issue that seemed somewhat similar to mine : [GitHub Issue](https://github.com/GIScience/openrouteservice/issues/339)

### Proposed solution

I dont have the necessary knowledge to be able to propose an actual solution, but i suspect it will probably be related to the way ORS handles OSM's tags and the weighs used when calculating routes?

### Additional context

_No response_

### Forum Topic Link

https://ask.openrouteservice.org/t/problematic-routing-from-ors/7089

aoles:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes #1981.

- Key functionality added: Introduce a weight penalty on roads tagged with `access = customers` in order to avoid them during route calculation.
- Reason for change:  sometimes the route returned by ORS takes "shortcuts" through parking lots, gas stations, or similar instead of following the regular road. This is of an issue because it:
  - might be illegal to do
  - is typically much slower due to the lower speed limit and  the presence of pedestrians
  - results in an overall higher number of turns compared to the direct route.

The following examples from the original issue now yield the desired routes.

image

image

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,8 +40,9 @@ Releasing is documented in RELEASE.md
 ### Removed
 
 ### Fixed
-- fix http pooling for matrix generation requests
-- bump deprecated CodeQL and failing Grype scan GitHub actions to their current versions
+- fix http pooling for matrix generation requests ([#2059](https://github.com/GIScience/openrouteservice/pull/2059))
+- bump deprecated CodeQL and failing Grype scan GitHub actions to their current versions ([#2061](https://github.com/GIScience/openrouteservice/pull/2061))
+- avoid routing on roads with access reserved to customers ([#2060](https://github.com/GIScience/openrouteservice/pull/2060))
 
 ### Security
 
--- a/ors-api/src/test/java/org/heigit/ors/apitests/routing/ResultTest.java
+++ b/ors-api/src/test/java/org/heigit/ors/apitests/routing/ResultTest.java
@@ -4598,7 +4598,7 @@ class ResultTest extends ServiceTest {
                 .then()
                 .assertThat()
                 .body("any { it.key == 'routes' }", is(true))
-                .body("routes[0].summary.distance", is(35.6f))
+                .body("routes[0].summary.distance", is(closeTo(35.6f, 1)))
                 .statusCode(200);
 
         coord2 = new JSONArray().put(8.682718).put(49.433239);
@@ -4614,7 +4614,7 @@ class ResultTest extends ServiceTest {
                 .then()
                 .assertThat()
                 .body("any { it.key == 'routes' }", is(true))
-                .body("routes[0].summary.distance", is(60.3f))
+                .body("routes[0].summary.distance", is(closeTo(60.3f, 1)))
                 .statusCode(200);
     }
 
--- a/ors-api/src/test/java/org/heigit/ors/apitests/routing/ResultTest.java
+++ b/ors-api/src/test/java/org/heigit/ors/apitests/routing/ResultTest.java
@@ -4579,6 +4579,45 @@ class ResultTest extends ServiceTest {
                 .statusCode(200);
     }
 
+    @Test
+    void testCustomersAccess() {
+        JSONArray coord1 = new JSONArray().put(8.682474).put(49.433011);
+        JSONArray coord2 = new JSONArray().put(8.682742).put(49.433153);
+        JSONArray coordinates = new JSONArray().put(coord1).put(coord2);
+
+        JSONObject body = new JSONObject()
+                .put("coordinates", coordinates)
+                .put("preference", "shortest");
+
+        given()
+                .headers(CommonHeaders.jsonContent)
+                .pathParam("profile", getParameter("carProfile"))
+                .body(body.toString())
+                .when()
+                .post(getEndPointPath() + "/{profile}")
+                .then()
+                .assertThat()
+                .body("any { it.key == 'routes' }", is(true))
+                .body("routes[0].summary.distance", is(35.6f))
+                .statusCode(200);
+
+        coord2 = new JSONArray().put(8.682718).put(49.433239);
+        coordinates = new JSONArray().put(coord1).put(coord2);
+        body.put("coordinates", coordinates);
+
+        given()
+                .headers(CommonHeaders.jsonContent)
+                .pathParam("profile", getParameter("carProfile"))
+                .body(body.toString())
+                .when()
+                .post(getEndPointPath() + "/{profile}")
+                .then()
+                .assertThat()
+                .body("any { it.key == 'routes' }", is(true))
+                .body("routes[0].summary.distance", is(60.3f))
+                .statusCode(200);
+    }
+
     private JSONArray constructBearings(String coordString) {
         JSONArray coordinates = new JSONArray();
         String[] coordPairs = coordString.split("\\|");
--- a/ors-api/src/test/java/org/heigit/ors/apitests/routing/ResultTest.java
+++ b/ors-api/src/test/java/org/heigit/ors/apitests/routing/ResultTest.java
@@ -4590,6 +4590,7 @@ class ResultTest extends ServiceTest {
                 .put("preference", "shortest");
 
         given()
+                .config(JSON_CONFIG_DOUBLE_NUMBERS)
                 .headers(CommonHeaders.jsonContent)
                 .pathParam("profile", getParameter("carProfile"))
                 .body(body.toString())
@@ -4598,7 +4599,7 @@ class ResultTest extends ServiceTest {
                 .then()
                 .assertThat()
                 .body("any { it.key == 'routes' }", is(true))
-                .body("routes[0].summary.distance", is(closeTo(35.6f, 1)))
+                .body("routes[0].summary.distance", is(closeTo(35.6, 1)))
                 .statusCode(200);
 
         coord2 = new JSONArray().put(8.682718).put(49.433239);
@@ -4606,6 +4607,7 @@ class ResultTest extends ServiceTest {
         body.put("coordinates", coordinates);
 
         given()
+                .config(JSON_CONFIG_DOUBLE_NUMBERS)
                 .headers(CommonHeaders.jsonContent)
                 .pathParam("profile", getParameter("carProfile"))
                 .body(body.toString())
@@ -4614,7 +4616,7 @@ class ResultTest extends ServiceTest {
                 .then()
                 .assertThat()
                 .body("any { it.key == 'routes' }", is(true))
-                .body("routes[0].summary.distance", is(closeTo(60.3f, 1)))
+                .body("routes[0].summary.distance", is(closeTo(60.3, 1)))
                 .statusCode(200);
     }
 
--- a/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/LimitedAccessWeighting.java
+++ b/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/LimitedAccessWeighting.java
@@ -27,12 +27,12 @@ import com.graphhopper.util.PMap;
  * @author Andrzej Oles
  */
 public class LimitedAccessWeighting extends AbstractAdjustedWeighting {
-    public static double DEFAULT_DESTINATION_FACTOR = 1;
-    public static double VEHICLE_DESTINATION_FACTOR = 10;
-    public static double DEFAULT_PRIVATE_FACTOR = 1.2;
-    public static double VEHICLE_PRIVATE_FACTOR = 10;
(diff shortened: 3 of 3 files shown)


ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/LimitedAccessWeighting.java
```diff
@@ -52,7 +58,9 @@ public LimitedAccessWeighting(Weighting superWeighting, PMap map) {
         destinationPenalty = checkBounds("road_access_destination_factor", map.getDouble("road_access_destination_factor", defaultDestinationFactor), MIN_DESTINATION_FACTOR, MAX_DESTINATION_FACTOR);
         double defaultPrivateFactor = encoder.getTransportationMode().isMotorVehicle() ? VEHICLE_PRIVATE_FACTOR : DEFAULT_PRIVATE_FACTOR;
         privatePenalty = checkBounds("road_access_private_factor", map.getDouble("road_access_private_factor", defaultPrivateFactor), MIN_PRIVATE_FACTOR, MAX_PRIVATE_FACTOR);
-        roadAccessEnc = destinationPenalty > 1 || privatePenalty > 1 ? encoder.getEnumEncodedValue(RoadAccess.KEY, RoadAccess.class) : null;
+        double defaultCustomersFactor = encoder.getTransportationMode().isMotorVehicle() ? VEHICLE_CUSTOMERS_FACTOR : DEFAULT_CUSTOMERS_FACTOR;
+        customersPenalty = checkBounds("road_access_customers_factor", map.getDouble("road_access_customers_factor", defaultCustomersFactor), MIN_CUSTOMERS_FACTOR, MAX_CUSTOMERS_FACTOR);
```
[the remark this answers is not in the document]
aoles: None of the `road_access_*` factors is currently subject to configuration.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/LimitedAccessWeighting.java
```diff
@@ -27,19 +27,25 @@
...
  */
 public class LimitedAccessWeighting extends AbstractAdjustedWeighting {
-    public static double DEFAULT_DESTINATION_FACTOR = 1;
-    public static double VEHICLE_DESTINATION_FACTOR = 10;
-    public static double DEFAULT_PRIVATE_FACTOR = 1.2;
-    public static double VEHICLE_PRIVATE_FACTOR = 10;
+    public static final double DEFAULT_DESTINATION_FACTOR = 1;
+    public static final double VEHICLE_DESTINATION_FACTOR = 10;
+    public static final double DEFAULT_PRIVATE_FACTOR = 1.2;
+    public static final double VEHICLE_PRIVATE_FACTOR = 10;
```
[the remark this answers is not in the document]
aoles: No need to do so as the names and code is self-explanatory.

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ Releasing is documented in RELEASE.md
 ### Removed
 
 ### Fixed
+- avoid routing on roads with access reserved to customers ([#2060](https://github.com/GIScience/openrouteservice/pull/2060))
```
sfendrich, line 45: Usually, we use the issue-id instead of the PR-id in the changelog.

.github/workflows/vulnerability-scanning.yml
sfendrich, line 1: How are the changes of these file related to the PR?

ors-api/src/test/java/org/heigit/ors/apitests/routing/ResultTest.java
```diff
@@ -4579,6 +4579,45 @@ void testBarriersAccessPermit() {
...
+        given()
+                .headers(CommonHeaders.jsonContent)
+                .pathParam("profile", getParameter("carProfile"))
+                .body(body.toString())
+                .when()
+                .post(getEndPointPath() + "/{profile}")
+                .then()
+                .assertThat()
+                .body("any { it.key == 'routes' }", is(true))
+                .body("routes[0].summary.distance", is(35.6f))
```
sfendrich: floating point values schould be compared with error margin.

ors-api/src/test/java/org/heigit/ors/apitests/routing/ResultTest.java
```diff
@@ -4579,6 +4579,45 @@ void testBarriersAccessPermit() {
...
+        given()
+                .headers(CommonHeaders.jsonContent)
+                .pathParam("profile", getParameter("carProfile"))
+                .body(body.toString())
+                .when()
+                .post(getEndPointPath() + "/{profile}")
+                .then()
+                .assertThat()
+                .body("any { it.key == 'routes' }", is(true))
+                .body("routes[0].summary.distance", is(60.3f))
```
sfendrich: floating point values should be compared with error margin

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/LimitedAccessWeighting.java
```diff
@@ -27,32 +27,38 @@
...
-    static double MIN_PRIVATE_FACTOR = 1;
-    static double MAX_PRIVATE_FACTOR = 10;
+    public static final double MIN_FACTOR = 1;// ensure that we do not need to change getMinWeight, i.e. road_access_factor >= 1
+    public static final double MAX_FACTOR = 10;
+    public static final double DEFAULT_DESTINATION_FACTOR = 1;
+    public static final double VEHICLE_DESTINATION_FACTOR = 10;
+    public static final double DEFAULT_PRIVATE_FACTOR = 1.2;
+    public static final double VEHICLE_PRIVATE_FACTOR = 10;
+    public static final double DEFAULT_CUSTOMERS_FACTOR = 1.2;
+    public static final double VEHICLE_CUSTOMERS_FACTOR = 1.5;
```
sfendrich, line 37: Shouldn't this be treated with similar values as access:destination?

sfendrich (changes requested):
Just a few minor comments.

CHANGELOG.md
aoles: Actually, I've seen both. While usually it doesn't make much of a difference, it still can happen that:
- although a PR fixes or implements the functionality described in an issue it actually goes beyond that description such that it is necessary to read through the PR's description in order to comprehend it fully 
- there is no corresponding issue to a PR
- a single PR tackles multiple issues

Therefore, I personally prefer to include the PR-id. Please note that PR ids is also what ends up in the auto-generated release notes.

.github/workflows/vulnerability-scanning.yml
aoles: Fair point 👍  I had to make these adjustments in order for the actions to run through, as they started failing for reasons seemingly unrelated to the contents of this PR.

Do you want me to open a separate PR addressing only the issues with the GitHub actions?

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/weighting/LimitedAccessWeighting.java
aoles: Not necessarily. If you set the factor as high as for `access = destination`, you start observing issues such as on reported in the original issue:

> And here is a somewhat ironic case where it tries very very hard to avoid going through the parking lot, even though it starts from within that very same parking lot:
[Peculiar routing](https://classic-maps.openrouteservice.org/directions?n1=45.639896&n2=-73.842764&n3=18&a=45.640016,-73.842491,45.640316,-73.843516&b=0&c=0&k1=en-US&k2=km)

In fact, it might make sense to loosen up the factors for destination a bit, but this is not the scope of this PR. I've played around with the factors and found that 1.5 already produces good results, while setting it too high might cause weird detours. As an example, consider the route to the gas station:

image

Once the factor is increased to 10 (the same as for destination) you get:

image

.github/workflows/vulnerability-scanning.yml
aoles: In fact, you can see these actions failing now on main after some recent merge.

sfendrich: O.K., then lets leave it in this PR

CHANGELOG.md
sfendrich: ok.

ors-api/src/test/java/org/heigit/ors/apitests/routing/ResultTest.java
aoles: Thanks for pointing this out! I've fixed it.

ors-api/src/test/java/org/heigit/ors/apitests/routing/ResultTest.java
aoles: Thanks for pointing this out! I've fixed it.

sfendrich (approved):
LGTM

PR #2075, opened 2025-05-30
fix: correct surface information for routes involving sidewalks

### LINKED ISSUE #2067: ORS returns wrong surface type for sidewalks

- [x] I have searched the existing issues

### Problem description

Way https://www.openstreetmap.org/way/314456049 is tagged with `sidewalk=right` and `sidewalk:right:surface=paving_stones` but routing with the walking profile returns asphalt as surface-type in the extra-infos:
https://classic-maps.openrouteservice.org/directions?n1=49.006985&n2=8.394571&n3=18&a=49.008105,8.394751,49.005874,8.394382&b=2&c=0&k1=en-US&k2=km

### Proposed solution

For pedestrian profiles extra infos tagged for the sidewalk should take precedence of the extra infos for the way.

### Additional context

_No response_

### Forum Topic Link

_No response_

aoles:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

Fixes #2067.
 
- Key functionality added: Process sidewalks attached to ways via `sidewalk = *` OSM tags into separate graph edges of `WayType.FOOTWAY` and take into account their surface as defined by `sidewalk:surface` or `sidewalk:*:surface` tags.

- Reason for change: Incorrect way type and surface returned in extra info in case of pedestrian routes that go along streets which have sidewalks attached to the OSM way representing the street.

The original example from the issue **before** ...

image

... and **after** applying the the changes in this PR:

image

### DIFF
--- a/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/WaySurfaceTypeGraphStorageBuilder.java
+++ b/ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/WaySurfaceTypeGraphStorageBuilder.java
@@ -18,6 +18,7 @@ import com.graphhopper.reader.ReaderWay;
 import com.graphhopper.storage.GraphExtension;
 import com.graphhopper.util.EdgeIteratorState;
 import lombok.Getter;
+import lombok.Setter;
 import org.heigit.ors.common.EncoderNameEnum;
 import org.heigit.ors.routing.graphhopper.extensions.SurfaceType;
 import org.heigit.ors.routing.graphhopper.extensions.WayType;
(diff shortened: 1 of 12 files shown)


ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/WaySurfaceTypeGraphStorageBuilder.java
```diff
@@ -17,21 +17,34 @@
 import com.graphhopper.reader.ReaderWay;
 import com.graphhopper.storage.GraphExtension;
 import com.graphhopper.util.EdgeIteratorState;
+import lombok.Getter;
```
sfendrich, line 20: Isn't lombok overkill for just one setter and one getter?

sfendrich (approved):
lgtm, thank you!

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/storages/builders/WaySurfaceTypeGraphStorageBuilder.java
aoles: In this case it might appear so. However, since we use lombok in the project anyway I would argue it makes sense to use it for any trivial getter and setter instead of implementing it.

PR #2135, opened 2025-09-25
refactor: transfer way category storage

sfendrich:
### Pull Request Checklist

         have been resolved.
         [Unreleased] heading.

- Key functionality added: Replaced WayCategoryStorage by EncodedValues


ors-api/src/test/resources/application-test.yml
```diff
@@ -99,6 +99,9 @@ ors:
               radius: 150
               output_log: false
               log_location: ors/traffic_log
+          encoded_values:
```
aoles: For the sake of backwards compatibility no `encoded_values` entries in the config should be necessary until the deprecation of extended storages has been completed and they are all removed  For now all the newly introduced EVs should be enabled based on the presence of `ext_storages.WayCategory`.

Please remove any new entries from the config file and instead make the necessary adaptations in the codebase.

ors-api/src/test/resources/application-test.yml
```diff
@@ -156,6 +159,9 @@ ors:
...
```
aoles: Same as above.

aoles (changes requested):
Thanks a lot! 🎉 

1. Please have a look at my comment regarding the changes to the config file. In general, we should refrain from introducing them otherwise we would need to do a major release which doesn't make sense unless the transition has been completed 🙈 

2. There are a few issues reported by SonarCube. I believe it's still worth to consider at least some of them in order to ensure high code quality standards 🙏 

3. Please address copilot's review comments 🤖 

Cheers,
Andrzej

ors-api/src/test/resources/application-test.yml
sfendrich: Done

ors-api/src/test/resources/application-test.yml
sfendrich: Done

sfendrich:
Thanks for the review! I've addressed all relevant issues. (To be honest yours were the only really relevant ones).
The Sonar issues are either invalid or not relevant for this PR. E.g., it does not make sense to refactor FastIsochrones, just because it uses a method that has changed. This would belong to a separate PR.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/util/parsers/FordParser.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public void createEncodedValues(EncodedValueLookup encodedValueLookup, List<EncodedValue> list) {
+        list.add(fordEnc);
+    }
+
+    @Override
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, boolean b, IntsRef intsRef1) {
+        //Keep in sync with documentation: waycategory.md
+
+        boolean isFord = way.getTag("ford", "").equals("yes");
```
aoles: You might consider using `way.hasTag("ford", "yes")` for the sake of a more readable and concise code.

pom.xml
```diff
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-parent</artifactId>
-        <version>3.4.9</version>
+        <version>3.4.10</version>
```
aoles, line 7: I think dependency updates shouldn't be smuggled in PRs focused on refactoring. Unless the refactoring actually requires the version bump please consider moving the update to a separate PR.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/AvoidFeaturesCoreEdgeFilter.java
```diff
@@ -18,31 +18,42 @@
...
+        tollwaysStorage = null;
     }
 
     @Override
     public final boolean accept(EdgeIteratorState iter) {
         if (iter instanceof RoutingCHEdgeIterator iterator && iterator.isShortcut())
             return true;
-        return (storage.getEdgeValue(iter.getEdge(), buffer) & avoidFeatures) == 0;
+        return formerWayCategory.accept(iter)
+            && acceptGenericTollways(iter);
```
aoles, line 50: The core is being build for a given profile. Therefore, I would expect that instead of accepting any tollway the core edge filter would take the profile into account by calling `tollwayExtractor.isProfileSpecificTollway` in a similar fashion as `AvoidFeaturesEdgeFilter` does in `acceptProfileSpecificTollways`. Or am I missing something here? 🤔

aoles (changes requested):
Thanks a lot for addressing my previous comments ❤️ 

Upon careful inspection I have a few more, please have a look below. Cheers!

pom.xml
sfendrich: Why not include a security fix?

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/AvoidFeaturesCoreEdgeFilter.java
sfendrich: This would introduce a change of behavior and should be discussed first.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/util/parsers/FordParser.java
sfendrich: done.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/AvoidFeaturesCoreEdgeFilter.java
aoles: So maybe it was wrong before. IMHO there is no use of e.g. including HGV tollways for car. In fact, this is even undesired as increased core size degrades the algorithm's performance.

pom.xml
aoles: Because any changes to dependencies could potentially cause disruptions on our end. The separation of changes to our code base from any external ones makes it easier to debug in case of any issues.

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/util/parsers/FordParser.java
aoles: Thanks! 👍

ors-engine/src/main/java/org/heigit/ors/routing/graphhopper/extensions/edgefilters/core/AvoidFeaturesCoreEdgeFilter.java
sfendrich: It's not so easy because the tollways categories overlap. Storing tollways profile-wise would make it more difficult to share a graph between profiles such as hgv & car. If we really want to change this, we should have a separate discussion.

pom.xml
sfendrich: Done. Apparently Github does not recognize when comments are outdated due to force-pushing.

aoles: Cheers! 👍

aoles (approved):
LGTM! 🚀

PR #2160, opened 2025-10-26
fix: docker compose image

ShadowArcanist:
A user reported that OpenRouteService was not working when deployed via Coolify.

The issue was traced to the Docker Compose file, which referenced the wrong image:
`local/openrouteservice:latest` instead of the correct public image `openrouteservice/openrouteservice:latest`.

image

This PR Fixes deployment issues for users following the Compose example (including on Coolify)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     ports:
       - "8080:8082"  # Expose the ORS API on port 8080
       - "9001:9001"  # Expose additional port for monitoring (optional)
-    image: local/openrouteservice:latest
+    image: openrouteservice/openrouteservice:latest
     # Advanced option! If you different ids to 0:0 and 1000:1000, you have to rebuild the container with the build args UID,GID.
     # The user command is useful if you want easier bind mount access or better security.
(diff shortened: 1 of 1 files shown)


aoles:
Hi [@ShadowArcanist](https://github.com/ShadowArcanist),

Thanks a lot for your suggestion 👍

I'm afraid there might be a misunderstanding on the intended usage of the bundled docker-compose file. This is meant for local usage when the repo is cloned and the image built locally, see [here](https://giscience.github.io/openrouteservice/run-instance/building-from-source#build-docker-image). In such a scenario it would be counterintuitive to pull `openrouteservice/openrouteservice:latest` which might be different from what the user has locally.

When deploying openrouteservice one would probably want to use a released version. Each release comes with release artifacts which include a dedicated docker-compose file where `local/openrouteservice:latest` is replaced by the specific version, e.g. [`openrouteservice/openrouteservice:v9.4.0`](https://github.com/GIScience/openrouteservice/releases/download/v9.4.0/docker-compose.yml). More details on using the released docker images can be found in the [documentation](https://giscience.github.io/openrouteservice/run-instance/running-with-docker#running-prebuilt-images).

Cheers,
Andrzej

ShadowArcanist:
Hi Andrzej, Thanks for clarifying, that makes perfect sense now. I appreciate the detailed explanation and the pointer to the release documentation!

PR #2215, opened 2025-12-17
ci: set the value of `docker_login` automatically

aoles:
Implement suggestion from Copilot's review 🚀


MichaelsJP (approved):
Nice. Good idea.

PR #2264, opened 2026-04-21
docs: add opencontainer metadata annotations to image

dklenke:
The Open Container Initiatives Image specification defines a set of optional annotations that add metadata to container images here https://github.com/opencontainers/image-spec/blob/main/annotations.md
Some of these annotations are used by downstream tools that scan images for various reasons. This PR adds 2 common annotations:
- **org.opencontainers.image.source** URL to get source code for building the image (string). Commonly used by dependency updater tools like [renovate](https://github.com/renovatebot/renovate) to find relevant releases and change logs.
- **org.opencontainers.image.licenses** License(s) under which contained software is distributed as an [SPDX License Expression](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Releasing is documented in RELEASE.md
 ## [unreleased]
 
 ### Added
+- add opencontainers metadata annotations to image ([#2264](https://github.com/GIScience/openrouteservice/pull/2264))
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


aoles (changes requested):
Thanks a lot for your contribution! ❤️ 

You might consider updating the CHANGELOG 😉 

Cheers!

aoles (approved):
LGTM 🚀

PR #2279, opened 2026-05-07
fix: base URL in OpenAPI specs

takb:
Reason for change:
- Update ORS public API base url in OAS generation code and docs
- Update default attribution string in default config


aoles (approved):
LGTM, Cheers! ❤️

PR #2280, opened 2026-05-08
fix: guard against invalid `mtb_scale` values

aoles:
During an attempt to store values that are outside of the encoder’s range, an uncaught exception is thrown, which causes the build process to fail. This fix prevents that from happening by ensuring the validity of the stored values.


takb (approved):
LGTM

PR #2292, opened 2026-06-09
fix: graphhopper upgrade barrier node handling

MichaelsJP:
### Pull Request Checklist

- [ ] 1. I have [**rebased**][rebase] the latest version of the main branch into my feature branch and all conflicts
         have been resolved.
         [Unreleased] heading.

Fixes # .

- Key functionality added:
- Reason for change:

-

-

### DIFF
--- a/ors-api/src/test/resources/application-test.yml
+++ b/ors-api/src/test/resources/application-test.yml
@@ -7,7 +7,7 @@ ors:
   endpoints:
     matrix:
       maximum_routes: 200
-      maximum_visited_nodes: 60000
+      maximum_visited_nodes: 65300
     isochrones:
       maximum_intervals: 10
     snap:
--- a/ors-engine/src/test/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReaderBarrierTest.java
+++ b/ors-engine/src/test/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReaderBarrierTest.java
@@ -1,0 +1,229 @@
+/*  This file is part of Openrouteservice.
+ *
+ *  Openrouteservice is free software; you can redistribute it and/or modify it under the terms of the
+ *  GNU Lesser General Public License as published by the Free Software Foundation; either version 2.1
+ *  of the License, or (at your option) any later version.
+
+ *  This library is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *  See the GNU Lesser General Public License for more details.
+
+ *  You should have received a copy of the GNU Lesser General Public License along with this library;
+ *  if not, see <https://www.gnu.org/licenses/>.
+ */
+package org.heigit.ors.routing.graphhopper.extensions;
+
+import com.graphhopper.reader.ReaderNode;
+import com.graphhopper.routing.OSMReaderConfig;
+import com.graphhopper.routing.ev.Subnetwork;
+import com.graphhopper.routing.util.EncodingManager;
+import com.graphhopper.storage.RAMDirectory;
+import com.graphhopper.util.PMap;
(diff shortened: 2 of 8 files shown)


ors-api/src/test/resources/application-test.yml
```diff
@@ -7,7 +7,7 @@ ors:
   endpoints:
     matrix:
       maximum_routes: 200
-      maximum_visited_nodes: 60000
+      maximum_visited_nodes: 100000
```
aoles: The value of `maximum_visited_nodes: 100000` is the default setting provided by `application.yml`, so redefining it here is unnecessary. I would suggest setting it to a lower value closer to the failure threshold such as `66000`, to catch any significant changes in graph topology.

ors-engine/src/test/java/org/heigit/ors/routing/graphhopper/extensions/ORSOSMReaderBarrierTest.java
```diff
@@ -0,0 +1,199 @@
...
+ *  See the GNU Lesser General Public License for more details.
+
+ *  You should have received a copy of the GNU Lesser General Public License along with this library;
+ *  if not, see <https://www.gnu.org/licenses/>.
+ */
+package org.heigit.ors.routing.graphhopper.extensions;
+
+import com.graphhopper.reader.ReaderNode;
+import com.graphhopper.routing.OSMReaderConfig;
+import com.graphhopper.routing.util.CarFlagEncoder;
```
aoles: The test runs against `com.graphhopper.routing.util.CarFlagEncoder` and `com.graphhopper.routing.util.FootFlagEncoder`, however, these are not the flag encoders ORS actually uses: `org.heigit.ors.routing.graphhopper.extensions.flagencoders.CarFlagEncoder` and `org.heigit.ors.routing.graphhopper.extensions.flagencoders.FootFlagEncoder`. It would be more reasonable to use the ORS counterparts here.

aoles (changes requested):
Thank a lot for the fix! 🎉 

Just some minor suggestions from the comments below, plus don't forget to update the CHANGELOG 😉

ors-api/src/test/java/org/heigit/ors/apitests/matrix/ResultTest.java
```diff
@@ -72,33 +72,22 @@ public ResultTest() {
         locations2.put(coord11);
         addParameter("locations2", locations2);
 
-        // Five new locations + all of the above ones
```
aoles, line 75: I am somewhat hesitant about this partial refactoring for the following reasons:

1. It introduces yet another way of defining the coordinates, in addition to the existing approach used for other `locations*`, increasing the code comprehension burden.
2. It adds redundancy by repeating the same coordinate values that were already defined.
3. Last but not least, it feels like a somewhat arbitrary addition that is not directly related to the primary goal of this PR, which is to fix an acute build issue.

Therefore, I believe that either the method of defining coordinates should be refactored consistently across the entire file, or not changed at all.

MichaelsJP: Thanks for the lower max visited nodes number. The new node handling pushes the limit and I just put it to 100k and added some more coordinates. It's technically correct, but not a necessary change. Agree.

 I figured out that, 65300 is the tightest we can set. The correct number now also allows us to remove the extra coordinates. Will revert it to the original state.

ors-api/src/test/java/org/heigit/ors/apitests/matrix/ResultTest.java
```diff
@@ -72,7 +72,6 @@ public ResultTest() {
...
```
aoles, line 75: I assume the comment has been deleted by mistake - please restore it.

aoles (changes requested):
Thanks for addressing my previous comments! Please see an additional remark below, and add your contribution to the CHANGELOG 😉

aoles (approved):
Perfect! ❤️

PR #2306, opened 2026-06-30
test: check elevation provider for preprocessed data

aoles:
Follow-up of https://github.com/GIScience/openrouteservice/pull/2305: add regression test.


takb (approved):
lgtm

PR #2319, opened 2026-07-23
fix: use direct mapping of node IDs to OSM IDs

### LINKED ISSUE #2315: Investigate file size differences with borders preprocessed
The graph files differ in size depending on whether the graph was built with borders preprocessed or not. This is only partially expected and needs to be investigated.

aoles:
OSM nodes can sometimes become duplicated, such as barrier nodes when creating zero-length edges. In these cases, the original reverse mapping (OSM IDs to node IDs) is no longer sufficient to correctly resolve node tags (such as "country") on a per-edge basis.

Fixes #2315.

**To be run with https://github.com/GIScience/graphhopper/pull/117**


jhaeu (approved):
LGTM and JL

jlink (approved):
Reviewed together with JH

takb:
Good find :+1:

